PR #5, opened 2017-09-11
feat(Card): Add missing card features

iamdavid0091:
1. Add static classes to all card components
2. Update styles
3. Add new card demos
4. Add support for vertical actions

### DIFF
--- a/src/components/Button.js
+++ b/src/components/Button.js
@@ -2,6 +2,10 @@ import styled, { css } from 'styled-components';
 import elevation, { elevationTransition } from '../mixins/elevation';
 import ripple from '../mixins/ripple';
 
+const ButtonComponent = props => (
+  <button className={`${props.className} smc-button`}>{props.children}</button>
+);
+
 const primary = css`
   color: ${props => props.theme.primary};
 `;
@@ -36,10 +40,10 @@ const raised = css`
   }
 `;
 
-const Button = styled.button`
+const Button = styled(ButtonComponent)`
   display: inline-block;
   position: relative;
-  min-width: 64px;
+  min-width: 88px;
   height: 36px;
   padding: 0 16px;
   border: none;
--- a/src/components/Card/Avatar.js
+++ b/src/components/Card/Avatar.js
@@ -1,6 +1,10 @@
 import styled from 'styled-components';
 
-const Avatar = styled.div`
+const AvatarComponent = props => (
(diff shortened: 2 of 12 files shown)


src/components/Button.js
```diff
@@ -2,6 +2,10 @@ import styled, { css } from 'styled-components';
 import elevation, { elevationTransition } from '../mixins/elevation';
 import ripple from '../mixins/ripple';
 
+const ButtonComponent = props => (
+  <button className={`${props.className} card-action`}>{props.children}</button>
```
brad-decker: A button component can exist outside of the card context so the static class name for button should not be card-action. It should instead be 'material-button' or something similar. smc-button (styled material components button?) if that were the adopted pattern we should also adopt smc-card

src/components/Card/Avatar.js
```diff
@@ -1,6 +1,10 @@
 import styled from 'styled-components';
 
-const Avatar = styled.div`
+const AvatarComponent = props => (
+  <div className={`${props.className} card-avatar`}>{props.children}</div>
```
brad-decker: For now this is fine but we may want to elevate the avatar outside of the card context to use in other components. At which time card-avatar would be refactored to smc(?)-avatar

src/components/Card/Card.js
```diff
@@ -1,7 +1,11 @@
 import styled from 'styled-components';
 import elevation from '../../mixins/elevation';
 
-const Card = styled.div`
+const CardComponent = props => (
+  <div className={`${props.className} demo-card`}>{props.children}</div>
```
brad-decker: Not demo card, but just card or smc-card

brad-decker:
@iamdavid0091 bumpity bump bump

src/components/Button.js
iamdavid0091: Done. Changed to `smc-card`.

src/components/Card/Avatar.js
iamdavid0091: Done. Changed to `smc-avatar`.

iamdavid0091: Done. Changed to `smc-avatar`.

src/components/Card/Card.js
iamdavid0091: Done. Changed to `smc-card`.

PR #7, opened 2017-09-21
feat(dialog): implement dialog styles

iamdavid0091:
implement most of the styles and components to support dialog creation


brad-decker:
Closing in favor of #83 which is now landed. Sorry @iamdavid0091

PR #57, opened 2017-11-22
Feat(TextField): add textfield

AriLFrankel:
Add a textfield component with the following props: hintText, autofocus, floating labels, fullWidth, helper/error text, validation.

This is a subset of the features that are required for Homepage MVP

[preview link](https://docs-nnmtjhdvzx.now.sh)


brad-decker:
Three things
1. 👏 👏 🗜 (typo'd but thought it was funny and kept it)
2. Can you update your styled components at the bottom of the main file so that they also have 'smc-xxxx' type static class names (future proofing)
3. Also build the static site and add the updated docs to your PR so the github pages for this repo update.

src/components/TextField.js
```diff
@@ -0,0 +1,109 @@
...
+class TextFieldComponent extends Component {
+  constructor(props) {
+    super(props);
+    this.state = {
+      text: props.defaultValue || '',
+      focus: false,
+    };
+  }
+
+  onChange = () => this.setState({ text: this.textInput.value })
```
brad-decker: you should probably check to see if the component was passed an onChange handle. If it is passed this item and it is a function, call it and pass the original event.

src/components/TextField.js
```diff
@@ -0,0 +1,109 @@
+import styled from 'styled-components';
+import React, { Component } from 'react';
+
+class TextFieldComponent extends Component {
+  constructor(props) {
+    super(props);
+    this.state = {
```
brad-decker: ES6 class properties for state instead of in the constructor unless there are other things that have to be done in the constructor.

src/components/TextField.js
```diff
@@ -0,0 +1,109 @@
...
+          display={!this.props.defaultValue && !this.state.text.length}>
+          {this.props.hintText}
+        </PlaceHolder>
+        <UnderlineFocus
+          focus={this.state.focus}
+          underlineColor={this.props.underlineColor}
+          underlineFocusColor={this.props.underlineFocusColor} />
+        <Input
+          value={this.state.text}
+          innerRef={(input) => {this.textInput = input;}}
```
brad-decker: I'm not opposed to this, but it might be easier to just use the event object that gets passed into onChange. event.target.value will hold the newly updated value. If we needed the this.textInput for the future or for other things i'd be keen on keeping it though.

src/components/TextField.js
```diff
@@ -0,0 +1,109 @@
...
+          autoComplete={false}
+          onChange={this.onChange}
+          onFocus={this.onFocus}
+          onBlur={this.onBlur}
+          className={'smc-text-field'} />
```
brad-decker: Does this syntax actually result in the className being applied? The reason i ask is my understanding was styled components set the className of the main element it wraps

src/components/TextField.js
```diff
@@ -0,0 +1,109 @@
...
+    );
+  }
+}
+
+
+const Container = styled.div`
+  font-size: 16px;
+  line-height: 24px;
+  width: ${(props) => {
+    if (props.fullWidth) {
```
brad-decker: Suh-weet

src/components/TextField.js
```diff
@@ -0,0 +1,109 @@
...
+  background-color: transparent;
+  font-family: Roboto, sans-serif;
+  cursor: auto;
+  border-bottom: 0.5px solid ${props => props.underlineColor || 'rgba(0,0,0,0.3)'}
+`;
+
+const PlaceHolder = styled.div`
+  position: absolute;
+  opacity: ${(props) => {
+    if (props.display) {
```
brad-decker:
```js
${props => + props.display}
```
unary operation + with true or false will yield 1 or 0, and has the added benefit of being obscure enough that people will ask what its doing. @Nathan-Schwartz @martahj am i being too creative here? We could also allow props => props.display ? 1 : 0

src/components/TextField.js
```diff
@@ -0,0 +1,109 @@
...
+  bottom: 0px;
+  width: 100%;
+`;
+
+const UnderlineFocus = styled.div`
+  position: absolute;
+  bottom: 0px;
+  border-top: 1.5px solid ${props =>
+    props.underlineFocusColor || props.underlineColor || props.theme.primary};
+  width: 0%;
```
brad-decker: could do
```js
width: ${props => props.focus ? 100 : 0}%;
```

src/components/TextField.js
```diff
@@ -0,0 +1,109 @@
...
+
+const Input = styled.input`
+  transition: all 1s;
+  padding: 0px;
+  position: relative;
+  width: 100%;
+  border: none;
+  outline: none;
+  background-color: rgba(0, 0, 0, 0);
+  color: rgba(0, 0, 0, 0.87);
```
brad-decker: Having these colors hardcoded is probably fine for our first iteration but we'll want to plug these into the theme manager context eventually so that they can be controlled on a global level.

src/components/TextField.js
```diff
@@ -0,0 +1,109 @@
...
+  cursor: inherit;
+  font-style: inherit;
```
brad-decker: were these added for a reason? Seems like this should just be the default

src/components/TextField.js
```diff
@@ -0,0 +1,109 @@
...
+
+
+const Container = styled.div`
+  font-size: 16px;
+  line-height: 24px;
+  width: ${(props) => {
+    if (props.fullWidth) {
+      return '100%';
+    }
+    return '256px';
```
martahj: I'm not against this as a default, but where does this value come from? The material design specs [don't specify](https://material.io/guidelines/components/text-fields.html#) and the material components library seems to show [167](https://material-components-web.appspot.com/text-field.html).

Also this might be a later enhancement, but we may want to allow the user to pass in a `width` prop to set the width to a larger or smaller value (though I guess they could do that through the `style` prop?)

src/components/TextField.js
martahj: Seconding this so the textfield can be controlled.

src/components/TextField.js
```diff
@@ -0,0 +1,109 @@
...
+          style={this.props.placeHolderStyle}
+          display={!this.props.defaultValue && !this.state.text.length}>
+          {this.props.hintText}
+        </PlaceHolder>
+        <UnderlineFocus
+          focus={this.state.focus}
+          underlineColor={this.props.underlineColor}
+          underlineFocusColor={this.props.underlineFocusColor} />
+        <Input
+          value={this.state.text}
```
martahj: If the parent passes a `value` prop, it should override this so the component can be controlled by the parent.

src/components/TextField.js
martahj: I agree with this

src/components/TextField.js
AriLFrankel: 👍

src/components/TextField.js
AriLFrankel: If I'm understanding you correctly this should be handled by the defaultValue prop

src/components/TextField.js
AriLFrankel: That's fair. I used this because I was exploring possibilities with focus, but I think it's not necessary at this stage. I can use event.target.value for now and refactor at a later time if a reference to the input becomes necessary.

src/components/TextField.js
AriLFrankel: This applies the classname to the input element. I wasn't sure about whether to apply at this level or the container. What are your thoughts on that, @brad-decker @martahj

src/components/TextField.js
AriLFrankel: merci

src/components/TextField.js
AriLFrankel: I took 256 from inspecting the examples on the docs page. Could u link me to where I can find 167?

Yes, they are able to pass in a width through the style prop

src/components/TextField.js
AriLFrankel: I like the unary operator cause it's cool looking.

Also, our linter doesn't like returning arrow functions and it's something I often find myself wanting to do, in this case included.

I went with the opacity because the display property isn't supported by animations. I want to animate the disappearance of hint text.

src/components/TextField.js
AriLFrankel: The default values on these fields are 'normal'

src/components/TextField.js
AriLFrankel: 👍

martahj: I don't think it is. Maybe I'm reading this component incorrectly, but it looks like the value is always set to `this.state.text`, which is updated whenever the user interacts with the textfield, and the role of `defaultValue` is to provide an initial value for `this.state.text`. This is different from a controlled component where the single source of truth for the value is held in the parent and passed in to the textfield.

Recall how we often use the Material UI textfield:

```js
class ParentComponent extends PureComponent{
  state = {
    value: this.props.property.pageTitle,
  };

  onChange = ({ target }) => this.setState({ value: target.value })

  render() {
    return (
      <Textfield
        onChange={this.handleChange}
        value={this.state.value}
      />
    );
  }
}
```

src/components/TextField.js
martahj: Just the first few examples on [this page](https://material-components-web.appspot.com/text-field.html).
screen shot 2017-11-22 at 3 00 14 pm

Could you link me to the docs you're looking at?

src/components/TextField.js
martahj: @AriLFrankel Got it, opacity decision makes sense 👍

src/components/TextField.js
brad-decker: I'm curious how it works cause Input already has a className defined by styled.input so its cool if passing a className into the component itself also concatenated the dynamic + static names together. 

I think that while the styled elements defined at the bottom of the file all exist only to serve this main component what we're going to find is that there is a need for these to be global (at least some of them) to be shared with other components (not necessary now, but i can see this happening down the road)... 

In that event its helpful to have a static classname for literally every thing we use for ease of descendant targeting.

src/components/TextField.js
```diff
@@ -0,0 +1,107 @@
...
+
+
+  onChange = (e) => {
+    if (this.props.onChange) {
+      return this.props.onChange(e);
+    }
+    this.setState({ text: e.target.value });
+  }
+
+  onFocus= () => this.setState({ focus: true })
```
brad-decker: You're gonna wanna do the same here that you did for this.props.onChange above to make sure the end user's onFocus and onBlur take effect.

src/components/TextField.js
```diff
@@ -0,0 +1,107 @@
...
+
+  onBlur = () => this.setState({ focus: false })
```
brad-decker: You're gonna wanna do the same here that you did for this.props.onChange above to make sure the end user's onFocus and onBlur take effect.

pages/text-fields.js
```diff
@@ -0,0 +1,68 @@
+import React, { PureComponent } from 'react';
+import styled from 'styled-components';
+import MaterialThemeProvider from '../src/theme/ThemeProvider';
+import TextField from '../src/components/TextField';
+
+class TextFieldPage extends PureComponent {
+  constructor() {
+    console.log('controlled value is ')
```
brad-decker: console.log

pages/text-fields.js
```diff
@@ -0,0 +1,68 @@
...
+    super();
+    this.state = {
```
brad-decker: es6 class properties for state

pages/text-fields.js
```diff
@@ -0,0 +1,68 @@
...
+      controlledInputValue: '',
+    };
+    this.handleChange = this.handleChange.bind(this);
```
brad-decker: no need to bind an es6 class property its already bound to the instance.

pages/text-fields.js
```diff
@@ -0,0 +1,68 @@
...
+  }
+
+  handleChange = (e) => {
+    console.log(`controlled value changing from ${this.state.controlledInputValue} to ${e.target.value}`);
```
brad-decker: console.log

src/components/TextField.js
brad-decker: can you point me to where these are set in the google material design spec and also what impact this has on the end result? By default css is inherited from their parent. If you set the font-size, weight, stretch, etc on a div, all elements in that div will start with that regardless of the 'default' value being normal.

src/components/TextField.js
```diff
@@ -0,0 +1,107 @@
...
+  transition: all 450ms;
+  bottom: 0px;
+  width: 100%;
+`;
+
+const UnderlineFocus = styled.div`
+  position: absolute;
+  bottom: 0px;
+  border-top: 1.5px solid ${props =>
+    props.underlineFocusColor || props.underlineColor || props.theme.primary};
```
brad-decker: why are we supporting two props for underlineFocusColor / underlineColor... whats the difference here?

brad-decker:
Almost there @AriLFrankel :)

src/components/TextField.js
AriLFrankel: Scratch that. It seems possible to do autofocus without this. I will take it out for now and can put it back in if a later feature seems to require it

src/components/TextField.js
AriLFrankel: It does seem to concatenate it. This input has both the className provided as a prop and the one generated by styledComponents.

That makes sense what you're saying about wanting a global reference to all parts of the component. In that case it might make sense to rename the above to smc-text-field-input and follow that pattern for components, such as smc-text-field-underline-focus, e.g.

src/components/TextField.js
AriLFrankel: This is not set in the material design spec. This is taken from the Material UI implementation.
I see us swapping out these hard coded styles to be taken from the theme.

As it turns out inputs don't default to inheriting their parent's styles, e.g. http://jsfiddle.net/rvjKE/

Taking them out the input is unstyled.

src/components/TextField.js
AriLFrankel: 👍

src/components/TextField.js
AriLFrankel: This is inspired by the MUI styled examples of text fields
http://www.material-ui.com/#/components/text-field

This allows users to specify a focus underline color without necessarily specifying a blur underline color

pages/text-fields.js
AriLFrankel: This was intentional as a demonstration of the controlled component. If it feels superfluous I can take it out

src/components/TextField.js
brad-decker: Does this exist on the material design spec? If it doesn’t then we need to remove. We aren’t copying material ui — that package is great but it’s not necessarily to spec

src/components/TextField.js
brad-decker: Again we arent using material-ui as inspiration.

AriLFrankel:
Changing this to a WIP for the moment while I remove features not included in Material Component spec and work on adding the rest that are.

src/components/TextField.js
```diff
@@ -0,0 +1,186 @@
...
+    this.props.onBlur && this.props.onBlur(e);
+    this.setState({ focus: false });
+  }
+
+  render() {
+    return (
+      <Container
+        error={this.state.error}
+        className={'smc-text-field-container'}
+        style={this.props.style}
```
martahj: This is a Material UI thing - I'm not super against keeping it, but it would be great to do this using `styled`.

AriLFrankel:
added link to PR, @martahj

martahj:
@iamdavid0091 @nataliaisabelle @jeanpage Could you take a look at the [preview](https://docs-nnmtjhdvzx.now.sh/) @AriLFrankel posted to make sure you're satisfied that the textfields meet material design?

nataliaisabelle:
@martahj @AriLFrankel Reviewing this now.

jeanpage:
@AriLFrankel I have an NIT on this 'a required field'. The asterisk remains red, but the rest of the hint text changes to gray on the focus state.

Also by Material Design's definition:
> _At the bottom of the form, mention that asterisks indicate required fields._

screen shot 2017-11-27 at 3 46 21 pm

nataliaisabelle:
@AriLFrankel Adding edits below, as I go through the fields:

### Updates:

**Floating Label (4th field):**
- [ ] Change floating label color onFocus of the field to the same color as the blue line. See screenshot below:
screen shot 2017-11-27 at 4 23 25 pm

AriLFrankel:
@jeanpage That is inspired by the implementation here http://material-components-web.appspot.com/text-field.html

If it's not what we want to do then I can take that out

jeanpage:
@AriLFrankel I was using this as my reference:
https://material.io/guidelines/components/text-fields.html#text-fields-layout

nataliaisabelle:
@AriLFrankel Take a look at my comment above. I think I got most of the edits. Let me know if you have any questions! Thanks.

AriLFrankel:
@jeanpage Got ya.

@nataliaisabelle @jeanpage Sounds like we want to use [material.io](https://material.io/guidelines/components/text-fields.html#text-fields-layout) as the ultimate source of truth and [material components](http://material-components-web.appspot.com/text-field.html) as a backup reference.

src/components/TextField.js
AriLFrankel: As per offline conversation, what we'll do is expose a prop to for styling each component of the text field. On this case, containerStyle will be included in the _styled_ styles with higher priority

AriLFrankel:
@jeanpage @nataliaisabelle [here](https://docs-ysfevhifds.now.sh) is an updated link
You'll still find the issue that @nataliaisabelle and I discussed about the baseline not being aligned with respect to character descent. That looks sort of tricky and may be something that I have to ask to revisit in the interest of getting an MVP through

@martahj This includes the change we discussed to style props

src/components/TextField.js
```diff
@@ -0,0 +1,224 @@
...
+          onChange={this.onChange}
+          onFocus={this.onFocus}
+          onBlur={this.onBlur}
+          className={'smc-text-field-input'} />
+      </Container>
+    );
+  }
+}
+
+const primaryTextColor = css`${props => props.theme.textColors.primary}`;
```
brad-decker: interesting idea defining these in this way. Not for this PR but we could actually create a file that exports these types for all of our other components that need theme definitions to make it super simple to understand how to use the theming!

martahj:
This will be ready to merge once @nataliaisabelle and @jeanpage sign off

nataliaisabelle:
@AriLFrankel Got it! Thanks. A few last edits:

- [ ] **When the field is idle and filled the floating label should be the 54% black.** 

Update on the following fields:
- Floating label
- Floating label with hint text
- Required field

_Desired:_

screen shot 2017-11-28 at 10 14 02 am

nataliaisabelle:
Looks good to me! @AriLFrankel Thanks!

jeanpage:
@AriLFrankel It looks like the padding and margin issues @natalia pointed out above are the same here. Did you adjust the padding around in the text fields?

AriLFrankel:
@jeanpage This has to do with the baseline not being aligned with respect to character descent. That looks sort of tricky and may be something that I have to ask to revisit in the interest of getting an MVP through.

nataliaisabelle:
@jean @AriLFrankel I think what we have here works for MVP. We can keep that item unchecked for now and revisit it when we have more time.

jeanpage:
Thanks @nataliaisabelle LGTM!

PR #61, opened 2017-11-28
build(package.json): Fix typo

martahj:
Fixes a typo in the 'clear' script in the package.json preventing the
package from being published.

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,7 +5,7 @@
   "main": "dist/index.js",
   "scripts": {
     "test": "yarn lint",
-    "clear": "rm -r dist/*",
+    "clear": "rm -rf dist/*",
     "build": "yarn clear && babel --out-dir dist src",
     "lint": "eslint src/**",
     "dev": "next dev -p 8080",


AriLFrankel:
I think that this is an ok change to make, but using rm -r is not a problem either as far as I understand.

https://www.computerhope.com/unix/urm.htm

martahj:
It was throwing an error for me, and this fixed it. It might be because `-r-` [prompts the user](https://www.computerhope.com/unix/urm.htm) for confirmation, and the script never gave it?

AriLFrankel:
hmm... I don't get this error. The prompt should only be for write-protected files.

What was the error?

martahj:
Looking back through logs, it looks like it was `rm: dist/*: No such file or directory`. So probably actually I hadn't built `dist` before and `-r` had a problem with that. Going to google to see if that's what it was or if something else is going on.

martahj:
@AriLFrankel Actually just did a simple experiment. It does indeed look like `-r` errors if the file doesn't exist while `-rf` does not.

screen shot 2017-11-28 at 4 15 59 pm

AriLFrankel:
Yes, I had the same issue when I didn't build first. nice, rm -rf would be a good fix then!

martahj:
I'm going to fix the commit message on this PR. I'd assumed it was a typo, but to @AriLFrankel 's point it's a valid command, just overly scrupulous  about the folder it's trying to delete existing.

martahj:
Fixed commit message

PR #65, opened 2017-12-04
Fix button

martahj:
* Imports react so the button component will not error on being imported in a third party library.
* Bumps the version
* Prettier changes

### DIFF
--- a/src/components/Button.js
+++ b/src/components/Button.js
@@ -1,7 +1,7 @@
-import React from "react";
-import styled, { css } from "styled-components";
-import elevation, { elevationTransition } from "../mixins/elevation";
-import ripple from "../mixins/ripple";
+import React from 'react';
+import styled, { css } from 'styled-components';
+import elevation, { elevationTransition } from '../mixins/elevation';
+import ripple from '../mixins/ripple';
 
 const ButtonComponent = props => (
   <button className={`${props.className} smc-button`}>{props.children}</button>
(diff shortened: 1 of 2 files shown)


src/components/Button.js
```diff
@@ -19,23 +20,26 @@ const raised = css`
...
+    props.accent &&
+    `
     color: white;
     background-color: ${props.theme.accent};
     &:before {
       color: black;
     }
-  `}
-  ${props => props.primary && `
+  `} ${props =>
```
brad-decker: this looks really weird. I assume this is a prettier thing but i dislike this format a lot.

src/components/Button.js
```diff
@@ -61,9 +65,7 @@ const Button = styled(ButtonComponent)`
   user-select: none;
   box-sizing: border-box;
   -webkit-appearance: none;
-  ${elevationTransition}
-
-  &:active {
+  ${elevationTransition} &:active {
```
AriLFrankel: Are the &active styles being applied without a semicolon after elevationTransition?

src/components/Button.js
```diff
@@ -19,23 +20,26 @@ const raised = css`
   &:active {
     ${elevation(8)};
   }
-  ${props => props.accent && `
+  ${props =>
+    props.accent &&
+    `
```
brad-decker: Also don't like this... i'm not sure what rule this is that is doing that but the line as it was written is fine.

src/components/Button.js
```diff
@@ -76,23 +78,22 @@ const Button = styled(ButtonComponent)`
...
   }
 
-  ${ripple()}
-
-  ${props => props.accent && accent}
-  ${props => props.primary && primary}
-  ${props => props.raised && raised}
-  ${props => props.compact && `padding: 0 8px;`}
-  ${props => props.dense && `
+  ${ripple()} ${props => props.accent && accent} ${props =>
```
brad-decker: what?

martahj: Hmm.. you're right, this doesn't seem right.

src/components/Button.js
```diff
@@ -19,23 +20,26 @@ const raised = css`
...
+      props.primary &&
+      `
     color: white;
     background-color: ${props.theme.primary};
     &:before {
       color: black;
     }
-  `}
-  fieldset:disabled &, &:disabled {
+  `} fieldset:disabled &,
```
AriLFrankel: Is this style being applied?

martahj:
We don't have prettier in this repo - the changes are coming from the "fix on save" feature of my editor.

brad-decker:
It has to be hooked into prettier somehow because maybe you have it installed globally and fixOnSave taps into it?

martahj:
@brad-decker @AriLFrankel Got rid of horrible prettier formatting

PR #67, opened 2017-12-06
Fixed Tabs First Pass

kristajg:
Now link: https://docs-cgwnlqgxew.now.sh

This is a first pass at the Tabs component. Focuses on Fixed Tabs (non scrollable) only for MVP.  Includes defaults & options for the following styles:
- activeBackgroundColor - hexa or rgb string
- passiveBackgroundColor - hexa or rgb string
- activeFontColor - hexa or rgb string
- passiveFontColor - hexa or rgb string
- showInkbar - default is true, boolean

Known TODO's:
- animations (inkbar && ripple)
- accept icon as option with or instead of text label
- accept onChange as props
- (future) scrollable tabs


src/components/Tabs/Tabs.js
```diff
@@ -0,0 +1,80 @@
...
+  text-transform: uppercase;
+  width: 100%;
+  color: ${props => props.theme.textColors.primary};
+`;
+
+const Content = styled.div`
+  font-family: lato, sans-serif;
+  text-transform: none;
+  display: block;
+  margin: 0 auto;
```
AriLFrankel: same question

src/components/Tabs/Tabs.js
```diff
@@ -0,0 +1,80 @@
...
+`;
+
+const getTabWidth = (width, numTabs) => width || window.screen.width / Number(numTabs);
```
AriLFrankel: I don't think we have access to window doing SSR, so we may want to move this logic into didMount.

https://github.com/zeit/next.js/wiki/FAQ

src/components/Tabs/Tabs.js
```diff
@@ -0,0 +1,80 @@
...
+  margin: 0 auto;
+  text-transform: uppercase;
+  width: 100%;
+  color: ${props => props.theme.textColors.primary};
+`;
+
+const Content = styled.div`
+  font-family: lato, sans-serif;
+  text-transform: none;
+  display: block;
```
AriLFrankel: why's this necessary?

src/components/Tabs/Tabs.js
kristajg: Next does SSR in dev mode, correct? I do have access to window while developing locally - will I not after building for production?

src/components/Tabs/Tabs.js
kristajg: There's probably some lines of style that aren't necessary - I pulled styles from material web component's examples in the browser then modified them

pages/tabs.js
```diff
@@ -0,0 +1,65 @@
+import React from 'react';
+import Tabs from '../src/components/Tabs/Tabs';
+import Tab from '../src/components/Tabs/Tab';
+import MaterialThemeProvider from '../src/theme/ThemeProvider';
+
+// TODO: animations (inkbar && ripple)
+// TODO: accept icon as option with or instead of text label
+// TODO: accept onChange as props
+// TODO: (future) scrollable tabs
```
martahj: Other possible TODO: controlled tabs

src/components/Tabs/Tab.js
```diff
@@ -0,0 +1,63 @@
...
+    activeBackgroundColor={activeBackgroundColor}
+    passiveBackgroundColor={passiveBackgroundColor}
+    activeFontColor={activeFontColor}
+    passiveFontColor={passiveFontColor}>
+    {icon || null}
+    {label}
+  </TabContainer>
+);
+
+export default Tab;
```
martahj: In general, we're moving to a pattern wherein we try to export styled-components (instead of react components rendering styled-components) in order to maintain a consistent styled-components api for the user of the library. See [this PR](https://github.com/ConciergeAuctions/styled-material-components/pull/68) for more detail on that conversation.

Could you refactor at least the components that will be importable on the client to follow this pattern?  You can check out the Button or Card component for an example. The TextField (although not all of its children) is also refactored to use this pattern in the PR linked above.

src/components/Tabs/Tabs.js
```diff
@@ -0,0 +1,89 @@
...
+    Children.forEach(props.children, (tab) => {
+      if (isValidElement(tab)) {
+        tabs.push(tab);
+      }
+    });
+    const numTabs = tabs.length;
+    this.state = {
+      selectedIndex: 0,
+      tabs,
+      numTabs,
```
martahj: Not a huge deal right now because `tabs` doesn't change, but `this.state.tabs.length` is fairly inexpensive to reference, so separating out numTabs doesn't give us that much but could create conflicting sources of truth if down the road anything was introduced that allowed the tabs to change.

src/components/Tabs/Tabs.js
```diff
@@ -0,0 +1,89 @@
...
+    };
+  }
+
+  componentDidMount() {
+    const tabWidth =
+      this.props.width ||
+      parseInt(this.tabs.getBoundingClientRect().width / this.state.numTabs, 10);
+    this.setState({ tabWidth }); // eslint-disable-line
```
martahj: Maybe shouldn't hold up this PR, but in a responsive environment the width of `this.tabs` could change after a resize event.

src/components/Tabs/Tabs.js
```diff
@@ -0,0 +1,89 @@
...
+    const tabContent = [];
+    const tabNavigation = this.state.tabs.map((tab, i) => {
+      const selected = this.getSelected(i);
+      tabContent.push(tab.props.children ? <Content>{tab.props.children}</Content> : undefined);
+      return cloneElement(tab, {
+        key: i,
+        index: i,
+        selected,
+        tabWidth: this.state.tabWidth,
+        numTabs: this.state.numTabs,
```
martahj: What does this component do with props such as `numTabs` and `tabWidth`?

nataliaisabelle:
@kristajg Looks good! The only thing I noticed is the following:

- [ ] Ink bar extends past the width of the tab:
screen shot 2017-12-06 at 11 47 00 am

pages/tabs.js
kristajg: What are controlled tabs?

martahj: Tabs whose active value is controlled by the parent component (analogous to controlled text fields, for example)

src/components/Tabs/Tabs.js
kristajg: Not a problem, can fix it in this one

src/components/Tabs/Tabs.js
kristajg: I'll remove unneeded styles

src/components/Tabs/Tabs.js
kristajg: same same

src/components/Tabs/Tabs.js
kristajg: It doesn't need numTabs (will remove) but it needs tabWidth for style props

kristajg:
@nataliaisabelle @martahj @AriLFrankel Updated the code and the testing link: https://docs-cgwnlqgxew.now.sh

PR #68, opened 2017-12-06
Prefix and suffix style

martahj:
Allows the user to customize the styles of the textfield prefix
and suffix. Also refactors the TextField to export styled-component
to facilitate this, as well as bumps the version.

Preview app here: https://docs-bdkcgekkev.now.sh/


brad-decker:
I'm not a fan of adding the props to target sub styles. All components exported from this library should be a styled component. Meaning all of them should allow for .extend. If all other components also have a static className, i think the api we should support for styling sub items is ..

```js
import { TextField } from 'textfield/import/path';

TextField.extend`
  & .smc-textfield-suffix {
    color: 'maroon';
  }
`;
```

Thoughts @martahj ?

martahj:
@brad-decker I think I agree with the overall sentiment, but I'd need to think about it some more. In particular, as mentioned in [this PR](https://github.com/ConciergeAuctions/Merlin/pull/64), `.extend` is buggy in nextjs, so I'd like to make sure there's an alternative approach that could be used until the `.extend` bug got fixed.

I think this conversation may deserve a separate issue for further discussion. The general pattern has already undergone code review, been accepted, and gotten merged. This PR simply fixes an issue with the implementation, so I'd rather not hold it up for a wholesale refactor of the component.

brad-decker:
@martahj i hear you, and I apologize for that. If i was being more thorough in other PRs i would have caught it and suggested this pattern then. I want to carefully monitor this package as it is meant to be a global solution that relies **heavily** on existing styled components apis. .extend is one of those. In particular the issue you pointed out with nextjs does not impact this open source project. I want to make sure that this doesn't become a version of material ui with sub style props. that was never the vision of this project.

In some scenarios looking at previously approved patterns is fine, but in this case I am pretty firm on not creating subStyle props that bloat the number of props required on our elements. I can create an issue to track against refactoring the existing substyle props. 

I will research the next.js issue further in our product and come up with a solution that will support this pattern.

brad-decker:
Also i don't think that the suggested fix will require a huge amount of effort in this PR to implement. Correct me if i'm wrong.

martahj:
If we want to be able to use the suffix/prefix style extension, we will need to figure out a way to get around the `.extend` issue before we can implement the suggested fix to it. Once that's been identified, the change is likely to be straightforward.

What are your thoughts on the next steps in resolving this issue? @brad-decker

brad-decker:
Go forward with work that will make the suggested api work on this repo, not in our product, and i will work on the .extend solution.

martahj:
@brad-decker @AriLFrankel Updated this PR to address the above discussion

PR #70, opened 2017-12-06
feat(Button): Passes props

dekkofilms:
I passed down the props, so that we could fire onClick events, and
other button props in the future.

I thought it would be beneficial to pass all the props through to the `button`? Is this a good idea?


dekkofilms:
@brad-decker good point. I can definitely safeguard it to an approved list of props.

PR #73, opened 2017-12-07
feat(Button): Safeguards props

dekkofilms:
This allows props to be safeguarded by an acceptableProps object.


kristajg:
@dekkofilms Hey can you version bump plz

PR #77, opened 2017-12-07
feat(Tabs): Style Nav

kristajg:
This adds the ability to style the tab navigation
background to a speicfied color. Also, the port
is updated to 8081 for internal convenience.

### DIFF
--- a/pages/tabs.js
+++ b/pages/tabs.js
@@ -41,6 +41,7 @@ const TabsPage = ({ className }) => (
         <li>activeFontColor - hexa or rgb string</li>
         <li>passiveFontColor - hexa or rgb string</li>
         <li>showInkbar - default is true, boolean</li>
+        <li>inkbarColor - default is accent theme color, or prop passed as a hexa or rgb string</li>
       </ul>
       <Tabs
         width='130'
--- a/src/components/Tabs/InkBar.js
+++ b/src/components/Tabs/InkBar.js
@@ -7,7 +7,7 @@ const InkBarContainer = styled.div`
   left: ${props => props.selectedIndex * props.tabWidth}px;
   width: ${props => props.tabWidth}px;
   height: 2px;
-  background-color: ${props => props.inkbarColor};
+  background-color: ${props => props.inkbarColor || props.theme.accent};
 `;
 
 const InkBar = ({ numTabs, selectedIndex, tabWidth, inkbarColor }) => (
--- a/src/components/Tabs/Tabs.js
+++ b/src/components/Tabs/Tabs.js
@@ -15,11 +15,14 @@ const Content = styled.div`
   text-transform: none;
 `;
 
+const TabNavigationContainer = styled.div`
+  background-color: transparent;
+`;
+
 class TabsComponent extends PureComponent {
   static defaultProps = {
     width: undefined,
     showInkbar: true,
-    inkbarColor: '#000',
   }
 
   constructor(props) {
@@ -79,8 +82,10 @@ class TabsComponent extends PureComponent {
 
     return (
       // eslint-disable-next-line
-      <TabsContainer innerRef={node => this.tabs = node} className={'smc-tab-bar'}>
-        {tabNavigation}
+      <TabsContainer innerRef={node => this.tabs = node} className='smc-tab-bar'>
+        <TabNavigationContainer className='smc-tab-navigation-container'>
+          {tabNavigation}
+        </TabNavigationContainer>
         {this.props.showInkbar ?
           <InkBar
             numTabs={this.state.numTabs}
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "styled-material-components",
-  "version": "0.0.18",
+  "version": "0.0.19",
   "description": "coming soon",
   "main": "dist/index.js",
   "scripts": {
@@ -8,7 +8,7 @@
     "clear": "rm -rf dist/*",
     "build": "yarn clear && babel --out-dir dist src",
     "lint": "eslint src/**",
-    "dev": "next dev -p 8080",
+    "dev": "next dev -p 8081",
     "export": "next build && next export -o docs",
     "prepublish": "yarn test && yarn build",
     "now-build": "yarn export && cd docs && now"


martahj:
nvm good to go after rebase

AriLFrankel:
What about a pattern of exposing a generalised style prop for each of the parts of the tabs component, in this case adding a tabStyle prop that we could use to pass in styles to a tab beyond just background color.

brad-decker:
I want to avoid style props similar to material ui...

Instead use the extend api and static class names on sub components to update styles of sub components

kristajg:
@brad-decker It's really difficult to extend the Tab component for custom styling in this use case due to how it's nested inside the Tabs component.

kristajg:
According to the designs on the homepage there are tabs on the search area that have a tab width with a transparent background, then underneath the Property Listings tabs have a white background that is the full-width of the screen

src/components/Tabs/Tabs.js
```diff
@@ -80,7 +85,9 @@ class TabsComponent extends PureComponent {
     return (
       // eslint-disable-next-line
       <TabsContainer innerRef={node => this.tabs = node} className={'smc-tab-bar'}>
-        {tabNavigation}
+        <TabNavigationContainer tabNavigationBackground={this.props.tabNavigationBackground}>
```
AriLFrankel: How about adding a static classname here that we can use to style this container? (echoing brad's sentiment)

brad-decker:
@kristajg can you give me more details on why its hard to do with the .extend api? if the tabs have a class name it should be possible. I am probably not close enough to see the issue so if you could explain why it won't work or possibly provide me a code example that attempts to use .extend that would fail i'd appreciate it.

kristajg:
@brad-decker It's because we need to have a styling for the wrapper around all of the tabs (navigation) since the tab nav & the content are separated out in the component return statement. I guess it could be accessed via extending the Tabs component & giving a class to the wrapper then it would look something like this:
```
const  CustomTabs = Tabs.extend`
  .tab-navigation-wrapper {
    background-color: blue;
  }
`;
```
But it wouldn't be very intuitive from a dev perspective
@AriLFrankel thoughts?

AriLFrankel:
I like the idea of having a uniquely identifying class on as many parts of the tabs component as possible, including the tabs-wrapper

kristajg:
@AriLFrankel Sounds good, I'll update w/that now 👍

brad-decker:
@kristajg - i think in our documentation we can address the various components that can be styled via the extend operation. I think this will take shape over time and we might implement a common syntax for class names so that its intuitive.

brad-decker:
your commit to remove the inkbar theme option went too far the other way, I think having theme colors used as defaults makes sense and can hit a majority of use cases. It could still be extended even if this is set.

@kristajg

kristajg:
@brad-decker I have been documenting what styles are accessible in the `pages/tabs.js` file, but perhaps it could be easier via the README. I thought that made sense to remove the Inkbar color styling prop since the Inkbar class could be reached via extending as well. I'm trying to keep it consistent but I can add it back in

kristajg:
Added InkBar background-color props back in @brad-decker @AriLFrankel

AriLFrankel:
Instead of having a default color of black on the inkbar, I would give it the _primary_ color off the theme to match the text of the tab as per the guideline provided [here](https://material.io/guidelines/components/tabs.html#tabs-specs) about matching tab indicator color to tab title font color.

kristajg:
@AriLFrankel according to the docs it says this (I'm assuming indicator is the InkBar):

Indicator
Height: 2dp
Color: #fff or accent color

This leads me to believe it should be the accent, not the primary

kristajg:
@ConciergeAuctions/engineers  Aware this needs squashed - wanted to make sure I didn't need a version bump before I do that? Thanks.

AriLFrankel:
@kristajg yes, I agree with you. I was looking at the mobile spec my mistake.

src/components/Tabs/Tabs.js
```diff
@@ -79,14 +82,16 @@ class TabsComponent extends PureComponent {
...
+        <TabNavigationContainer className='smc-tab-navigation-container'>
+          {tabNavigation}
+        </TabNavigationContainer>
         {this.props.showInkbar ?
           <InkBar
             numTabs={this.state.numTabs}
             selectedIndex={this.state.selectedIndex}
             tabWidth={this.state.tabWidth}
-            inkbarColor={this.props.inkbarColor} /> :
+            inkbarColor={this.props.inkbarColor ? this.props.inkbarColor : null} /> :
```
AriLFrankel: I think we could get away with just `this.props.inkbarColor`

If there is no prop passed down wouldn't a falsey value get passed down anyway

pages/tabs.js
```diff
@@ -41,6 +41,7 @@ const TabsPage = ({ className }) => (
         <li>activeFontColor - hexa or rgb string</li>
         <li>passiveFontColor - hexa or rgb string</li>
         <li>showInkbar - default is true, boolean</li>
+        <li>inkbarColor - default is primary theme color, hexa or rgb string</li>
```
AriLFrankel: inkbarColor would be accent, right? I think it would also be nice to specify the default values on the other props

src/components/Tabs/InkBar.js
```diff
@@ -7,7 +7,7 @@ const InkBarContainer = styled.div`
   left: ${props => props.selectedIndex * props.tabWidth}px;
   width: ${props => props.tabWidth}px;
   height: 2px;
-  background-color: ${props => props.inkbarColor};
+  background-color: ${props => (props.inkbarColor ? props.inkbarColor : props.theme.accent)};
```
AriLFrankel: NIT: This could be reduced to `props.inkbarColor || props.theme.accent`

martahj:
I'm good to go on this, pending updates based on a few of @AriLFrankel 's comments.

It will need a patch version bump. I think 0.0.19 is next.

kristajg:
Thanks @AriLFrankel & @martahj - will update now

pages/tabs.js
kristajg: right, forgot to update the docs here

src/components/Tabs/Tabs.js
kristajg: correct!

kristajg:
@martahj @AriLFrankel updated!

brad-decker:
LGTM thanks @kristajg

brad-decker:
@AriLFrankel can you re-review?

pages/tabs.js
```diff
@@ -41,6 +41,7 @@ const TabsPage = ({ className }) => (
         <li>activeFontColor - hexa or rgb string</li>
         <li>passiveFontColor - hexa or rgb string</li>
```
AriLFrankel: It might be good to add the default values for these other fields.

src/components/Tabs/Tabs.js
```diff
@@ -15,11 +15,14 @@ const Content = styled.div`
   text-transform: none;
 `;
 
+const TabNavigationContainer = styled.div`
+  background-color: transparent;
+`;
+
 class TabsComponent extends PureComponent {
   static defaultProps = {
     width: undefined,
```
AriLFrankel: Isn't this implicitly the default value of a prop not passed in?

kristajg: I think I added this bc an error occurred if it wasn't explicitly added in default props

AriLFrankel: 🤔  Seems strange. I tried running it locally without that default prop and it seems to work fine

brad-decker: Although true, it isn't specifically an issue. We can iterate on this and clean it up later. Gonna merge.

PR #79, opened 2017-12-25
Drawer Component + Portal first pass

### LINKED ISSUE #9: Make Avatar a customizable component
The Avatar is currently part of the Card component and should be broken out to be extended and used in other components

brad-decker:
Also fixes #9 

https://docs-txsaxitvca.now.sh


src/components/Drawer.js
```diff
@@ -0,0 +1,36 @@
...
+}) => (
+  <Portal
+    showOverlay={open}
+    onRequestClose={handleRequestClose}
+    renderContents={() => (
+      <Sidebar open={open}>
+        {children}
+      </Sidebar>
+    )} />
+);
```
martahj: It seems like you could make the Drawer itself a styled-component, maybe passing the `this.props.className` into the div inside portal (or I guess to the portal itself)

src/components/Avatar.js
```diff
@@ -0,0 +1,24 @@
...
+  height: 2.5rem;
+  width: 2.5rem;
+  border-radius: 50%;
+  overflow: hidden;
+
+  & .smc-avatar-img {
+    width: 100%;
+    display: block;
+  }
+`;
```
martahj: The avatar is not considered a [material component](https://material.io/guidelines/material-design/introduction.html). We discuss this problem a bit in [this issue](https://github.com/ConciergeAuctions/styled-material-components/issues/55) but I don't think we came to a definite conclusion on what to do with elements that are useful and reusable but which are not listed as components in the material spec. I'm very slightly leaning towards exporting them but calling them something other than "components", but don't have very strong feelings one way or another at this point.

pages/drawer.js
```diff
@@ -0,0 +1,26 @@
+import React, { Component, Fragment } from 'react';
```
martahj: Could you add a link to this page in `pages/index.js`?

pages/drawer.js
```diff
@@ -0,0 +1,26 @@
...
+import Button from '../src/components/Button';
+import MaterialThemeProvider from '../src/theme/ThemeProvider';
+import { Drawer } from '../src/components/Drawer';
+
+export default class PortalPage extends Component {
```
martahj: Is the idea that this `PortalPage` could be reused? If not, since other pages might use portals, it might make more sense to call it `DrawerPage` or something similar

src/components/Avatar.js
brad-decker: I don't see the harm in exporting them as components -- sure they might be part of material spec as stand alone components but they are useful in many places as parts of other components. It is my hope that this library, more than any other, will be usable to build NEW components that are material in nature but are outside of the spec.

src/components/Drawer.js
brad-decker: The drawer itself is a styled-component (Sidebar) The portal is just a way of rendering it in a specific place outside of the normal DOM flow.

pages/drawer.js
brad-decker: Oops my bad

pages/drawer.js
brad-decker: Good call.

brad-decker:
@martahj just pushed up a styling update and added in some state to the portal component to make it slightly more performant.

src/components/Avatar.js
AriLFrankel: I think that there is a lot of utility to exporting this component - it seems like an easy win for some added flexibility in the library.

I see the concern about deviating from Material spec but I share the sentiment that it would be great to facilitate new Material components by exporting constituent parts. 

My concern with creating a separate folder for supporting components that aren't Material recognized components is twofold:
1. My main concern is that we are linking our folder structure to that of the Material spec. If Avatar becomes a recognized Material component we'd have to change our folder structure to match the new pattern, moving the Avatar from supporting components folder to Components. this would introduce a breaking change to our API.
2. A minor concern is that this forces consumers of this library to be aware of what is a recognized Material component and what is not as they contribute to and consume the library. This is significantly mitigated by documentation.
Overall, I think there's less cognitive overhead if we keep all components in one folder.

pages/drawer.js
```diff
@@ -0,0 +1,26 @@
+import React, { Component, Fragment } from 'react';
+import Button from '../src/components/Button';
```
AriLFrankel: NIT: Looks like we're not using Button.

src/components/Drawer.js
```diff
@@ -0,0 +1,50 @@
...
+import elevation from '../mixins/elevation';
+import { Portal } from '../components/Portal';
+
+const BaseSidebar = props => (
+  <aside className={`${props.className} ${props.open ? 'open' : ''} smc-drawer`}>
+    {props.children}
+  </aside>
+);
+
+const Sidebar = styled(BaseSidebar)`
```
AriLFrankel: request: It could be nice to be able to pass in a size

pages/drawer.js
```diff
@@ -0,0 +1,26 @@
...
+    open: false,
+  };
+
+  toggleOpen = () => this.setState({ open: !this.state.open });
+
+  render() {
+    return (
+      <MaterialThemeProvider theme={{ primary: '#03A9F4' }}>
+        <Fragment>
+          <Drawer open={this.state.open} handleRequestClose={this.toggleOpen}>
```
AriLFrankel: How would I do a non-temporary drawer?

If I pass in `temporary={false}`, open still would default to false.

src/components/Drawer.js
brad-decker: Doesn't seem to be a thing in the spec as far as i can tell @AriLFrankel - its like set in stone at specific sizes based on view ports - i started looking into sizing but haven't landed on an approach yet :+1:

pages/drawer.js
brad-decker: Haven't gotten around to the permanent ones yet -- or rendering outside of a portal

brad-decker:
Closing in favor of #85

PR #81, opened 2017-12-31
Slider

martahj:
Adds the Slider component.

Documentation here: https://docs-wzbhnalgcb.now.sh


src/components/Slider/SliderTrack.js
```diff
@@ -0,0 +1,155 @@
...
+  }
+
+  enableFocus = () => this.setState({ focused: true });
+  disableFocus = () => this.setState({ focused: false });
+
+  enableTransitions = cb => this.setState({ useTransitions: true }, cb);
+  disableTransitions = cb => this.setState({ useTransitions: false }, cb);
+
+  render() {
+    const { value, max, min, disabled } = this.props;
```
AriLFrankel: NIT: seems like max doesn't get used

src/components/Slider/SliderTrack.js
```diff
@@ -0,0 +1,155 @@
...
+    const { pixelsPerValue, valuePerPixel, useTransitions, focused, width } = this.state;
```
AriLFrankel: NIT: nor useTransitions

src/components/Slider/Trackline.js
```diff
@@ -0,0 +1,26 @@
...
+import styled from 'styled-components';
+
+const TracklineComponent = (props) => (
+  <div className={`${props.className} smc-slider-track-line`}>
+    {props.children}
+  </div>
+);
+
+// TODO use focus color
+const getTracklineBackground = (props) => {
```
AriLFrankel: Why can't this be folded into the TrackLine styles?

src/components/Slider/Slider.js
```diff
@@ -0,0 +1,79 @@
...
+  constructor(props) {
+    super(props);
+    const { min, max, initialValue, step } = props;
+    const actualMin = typeof min === 'number' ? min : DEFAULT_MIN;
+    const actualMax = typeof max === 'number' ? max : DEFAULT_MAX;
+    const value = typeof initialValue === 'number'
+      ? initialValue
+      : getAverage(actualMin, actualMax);
+    const actualStep = step || DEFAULT_STEP;
+    this.state = {
```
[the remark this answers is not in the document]
martahj: Not really, no

src/components/Slider/Slider.js
```diff
@@ -0,0 +1,79 @@
...
+class SliderComponent extends PureComponent {
+  constructor(props) {
+    super(props);
+    const { min, max, initialValue, step } = props;
+    const actualMin = typeof min === 'number' ? min : DEFAULT_MIN;
+    const actualMax = typeof max === 'number' ? max : DEFAULT_MAX;
+    const value = typeof initialValue === 'number'
+      ? initialValue
+      : getAverage(actualMin, actualMax);
+    const actualStep = step || DEFAULT_STEP;
```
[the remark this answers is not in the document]
martahj: Good call

src/components/Slider/Slider.js
```diff
@@ -0,0 +1,79 @@
...
+    }
+    this.setState({ value: actualValue });
+  }
+
+  render() {
+    const { min, max, step } = this.state;
+    const { continuous } = this.props;
+    const propsValue = this.props.value;
+    const stateValue = this.state.value;
+    const valueFromCorrectSource = typeof propsValue === 'number' ? propsValue : stateValue;
```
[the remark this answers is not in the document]
martahj: I can add that functionality. I hadn't done much thinking about what we'd want to do if the value passed was illegitimate

src/components/Slider/Trackline.js
martahj: There's no reason it can't be - I just find it easier to look at multiline style functions if they're broken out for some reason

martahj:
@AriLFrankel Updated based on your comments

src/components/Slider/Slider.js
```diff
@@ -0,0 +1,84 @@
...
+    const { min, max, initialValue, step } = props;
+    const actualMin = typeof min === 'number' ? min : DEFAULT_MIN;
+    const actualMax = typeof max === 'number' ? max : DEFAULT_MAX;
+    if (actualMax <= actualMin) throw new Error(`Slider minimum ${actualMin} exceeds maximum ${actualMax}`);
+    const value = typeof initialValue === 'number'
+      ? initialValue
+      : getAverage(actualMin, actualMax);
+    if (value < actualMin) throw new Error(`Slider value ${value} is less than minimum ${actualMin}`);
+    if (value > actualMax) throw new Error(`Slider value ${value} exceeds maximum ${actualMax}`);
+    const actualStep = step || DEFAULT_STEP;
```
AriLFrankel: It seems like we would want to add type validation on actualStep as we are for actualMin and actualMax so that we catch invalid values as well as undefined.

src/components/Slider/Trackline.js
```diff
@@ -0,0 +1,26 @@
+import React from 'react';
+import styled from 'styled-components';
+
+const TracklineComponent = (props) => (
+  <div className={`${props.className} smc-slider-track-line`}>
+    {props.children}
+  </div>
+);
+
+// TODO use focus color
```
AriLFrankel: It seems like this is possible already, is this a matter of deciding on the correct focus color?

src/components/Slider/Slider.js
martahj: It will use `DEFAULT_STEP` if `step` is undefined, but I'll add a check that step is a nonzero number 👍

src/components/Slider/Trackline.js
martahj: In the spirit of taking a lean approach to this library, I didn't take the focus color to the finish line. It's not a huge extra lift but since it's just a slight darkening of the track when it's focused (I think iff we're using a light theme - haven't thought about if and how to do this if the track background is being manually overridden using `styled` instead of being set by the theme) I'm leaning towards leaving it unfinished for now over letting it hold up the PR.

AriLFrankel: that makes sense :)

src/components/Slider/Slider.js
AriLFrankel: yes, exactly, I was thinking of values being passed in that are not numbers > 0

martahj:
@AriLFrankel I added a check for if a step of 0 was passed. If passed a nonzero number less than 0, we can just use the absolute value.

brad-decker:
Looks like a rebase is in order and then merge.

brad-decker:
@martahj bump on rebase.

martahj:
Waiting on Taylor and Krista's PRs for this one

martahj:
rebased y'all

PR #82, opened 2018-01-03
Portal

brad-decker:
This also includes an update to the Avatar component to make it reusable + a syntax update to naming conventions


src/components/Avatar.js
```diff
@@ -0,0 +1,24 @@
...
+  <div className={`${props.className} smc-avatar`}>
+    {props.src ? (
+      <img src={props.src} className="smc-avatar-img" />
+    ) : props.children}
+  </div>
+);
+
+export const Avatar = styled(AvatarComponent)`
+  position: absolute;
+  background: #bdbdbd;
```
kristajg: Should this be a theme color? I'm fine with the neutral gray, just wondering

brad-decker: Probably a good thing to allow in the future with theming, I think this color was pulled from the spec as a hard coded thing (i didn't actually change this in this PR just moved the file which i guess shows it as an entirely new thing).

PR #84, opened 2018-01-04
feat(List): Add

dekkofilms:
Added in List component

Link: https://docs-prttnrkkat.now.sh/lists/


pages/lists.js
```diff
@@ -0,0 +1,86 @@
...
+          <ListItem>
+            <ListItemText primary="Primary Title" secondary="This is the secondary line" />
+          </ListItem>
+        </List>
+      </div>
+      <h1>List with more text</h1>
+      <div className="list">
+        <List>
+          <ListItem leftAvatar="http://www.wasiofaces.com/wp-content/uploads/San-Diego-Headshots-Photography-041.jpg">
+            <ListItemText primary="Primary Title" secondary={
```
martahj: Interesting.. what's the reason for having the user import and use both a `ListItem` and a `ListItemText`? It seems like they could potentially be folded into one component for easier use

dekkofilms: Yeah! So looking at some of the patterns out there, it might not seem beneficial to have those separate at this point, but in future when we add in like `ListItemIcon` and `ListItemSecondaryAction` and some of those other things ... I feel `ListItem` will help group these. Thoughts?

martahj: I agree it will help group them, and it's a nice meet-in-the-middle between the two approaches (a lot of components vs a lot of props).

Hmm.. if the ListItem is going to do any organization/styling my initial thought it that it would be more intuitive if the ListItem contained the portion that the list was *definitely* going to have (the primary text, and maybe the secondary text) and then users could optionally pass in other items like the avatar. Off the top of my head, it seems like there might be a couple pain points when it comes to coding that though..

Thoughts?

dekkofilms: Yeah, I think we could for sure do that. I did notice in Material-UIs stable version they have it like you suggest, and in the new beta Material-UI, they have it more so like it above. I'm wondering if they faced some pain points down the road.

I definitely can change it to where ListItem will contain primary and secondary text ... if you think that's the better approach.

My only other thought is that if it ever comes to pass in other components into `ListItem`, it feels at least more intuitive and easier to do that as children, but props could still happen as well. (i.e. maybe instead of having a `leftAvatar` prop, just having the `Avatar` component a child of the `ListItem`). 

Sorry I feel like I'm word vomiting now. I'm down for your suggestion if we are good to move forward with that!

martahj: I agree it feels more correct to pass other components in as children.

Tbh I'd be 👍  on just moving forward with your implementation, and seeing if it needs to be refactored based on how it ends up being used. I don't think it's worth refactoring for at this point.

dekkofilms: Awesome! Sounds good to me, I'm sure it will evolve in the future as additions happen.

brad-decker:
LGTM - hold for @kristajg 's pr to get merged and then rebase.

PR #85, opened 2018-01-04
Drawer Component

brad-decker:
[Demo](https://docs-gfzxmhoiwl.now.sh/)

Features:
1. Both Left and Right attachment modes work
2. Implements an Overlay for temporary drawers
3. Animations per material design spec @iamdavid0091 <3
4. implements both persistent and temporary drawers (not permanent!)
5. Persistent drawers achieved by cool body: flex, :before and :after manipulation. 
6. injectGlobal styles allows us to guarantee that someone using our library has some specific required styles on their site without requiring a ridiculous "getting started" guide. It also could be dangerous (hot reloading is a little wacky with portal but i plan to fix in a future PR)


AriLFrankel:
Can you update the link please?

brad-decker:
@AriLFrankel @martahj @kristajg @dekkofilms ready for review.

src/components/Portal/Overlay.js
```diff
@@ -0,0 +1,27 @@
...
+  right: 0;
+  bottom: 0;
+  top: 0;
+  opacity: ${props => (props.open ? 1 : 0)};
+  transition: opacity 0.3s 0ms cubic-bezier(0, 0, .2, 1);
+  will-change: opacity;
+  pointer-events: ${props => (props.open ? 'inherit' : 'none')};
+  contain: strict;
+  overflow: hidden;
+  z-index: 4;
```
AriLFrankel: curious how come z-index of 4 here?

src/components/Portal/Portal.js
```diff
@@ -1,38 +1,16 @@
 import React, { Component } from 'react';
 import { createPortal } from 'react-dom';
-import styled from 'styled-components';
+// import styled from 'styled-components';
```
AriLFrankel: did you want to leave this in?

src/components/Drawer.js
```diff
@@ -0,0 +1,114 @@
...
+  overflow: hidden;
+  position: absolute;
+  transition: transform .195s;
+  ${drawerSizing}
+  will-change: transform;
+  display: flex;
+  flex-direction: column;
+  box-sizing: border-box;
+  touch-action: none;
+  ${props => props.temporary && `z-index: 5`}
```
AriLFrankel: commented on the z-index of 4 in Portal/Overlay as well. Curious about this. Is the idea to leave some room in between for users of this component?

src/components/Portal/Portal.js
```diff
@@ -64,10 +42,17 @@ export class Portal extends Component {
 
   render() {
     if (!this.state.portalMounted) return null;
+    const PortalContainer = this.props.mode === 'overlay'
+      ? Overlay
+      : Shift;
```
martahj: This API could be a little confusing. It looks like, if the user does not want to use mode `'overlay'`, they should pass in any mode prop *except* overlay. It seems a lot more common to specify a prop the user should pass in to override the default behavior than to allow any prop except the default to override it.

If we're only expecting the two modes, they could optionally pass in a boolean `shift={true}` if they wanted to avoid the overlay behavior. That way the default would be for `this.props.shift` to be falsey, and lines 45-47 could be rewritten:
```js
const PortalContainer = this.props.shift ? Shift : Overlay;
```

src/components/Portal/Portal.js
```diff
@@ -64,10 +42,17 @@ export class Portal extends Component {
...
+
     return createPortal(
-      <Overlay showOverlay={this.props.showOverlay} onClick={this.props.onRequestClose}>
+      <PortalContainer
+        direction={this.props.attachment}
+        open={this.props.open}
+        onClick={this.props.onRequestClose}>
```
martahj: I'm a bit conflicted on whether we should pass down all these props regardless of whether the PortalContainer is an Overlay or a Shift. On one hand, it doesn't really hurt anything. On the other hand, it's a bit misleading the pass down props that we know won't be used by the component. What do you think?

src/components/Portal/Portal.js
brad-decker: :+1: yeah the text string is not the correct approach for this. Thank you.

src/components/Portal/Portal.js
brad-decker: I think until we are fully aware of what these two containers need it doesn't hurt to pass them all, esp. considering there is so few. If this evolves into something more complex we can filter props appropriately.

src/components/Drawer.js
brad-decker: z-index 5 for the temporary drawer is from material components implementation of temporary drawers

src/components/Portal/Overlay.js
brad-decker: it should be directly under the z-index of 5 for the drawer itself.

src/components/Portal/Portal.js
brad-decker: Nope!

brad-decker:
@martahj @AriLFrankel feedback should be addressed

pages/drawer.js
```diff
@@ -0,0 +1,72 @@
...
+  toggleTemporaryLeft = () => this.setState({ temporaryLeft: !this.state.temporaryLeft });
+  toggleTemporaryRight = () => this.setState({ temporaryRight: !this.state.temporaryRight });
+  togglePersistentLeft = () => this.setState({
+    persistentLeft: !this.state.persistentLeft,
+    persistentRight: false,
+  });
+  togglePersistentRight = () => this.setState({
+    persistentRight: !this.state.persistentRight,
+    persistentLeft: false,
+  });
```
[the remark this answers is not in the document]
brad-decker: Good idea. Need to get in the habit of doing this

pages/drawer.js
```diff
@@ -0,0 +1,79 @@
...
+    persistentLeft: false,
+    persistentRight: false,
+  };
+
+  toggleTemporaryLeft = () => this.setState(prevState => ({
+    temporaryLeft: !prevState.temporaryLeft,
+  }));
+
+  toggleTemporaryRight = () => this.setState(prevState => ({
+    temporaryLeft: !prevState.temporaryRight,
```
martahj: Should this be `temporaryRight: !prevState.temporaryRight` ?

brad-decker: ugh. copy paste has failed me

brad-decker:
@martahj addressed last question. :)

brad-decker:
Version bumped and rebased.

martahj:
@dekkofilms @AriLFrankel do you want to re-review this PR?

AriLFrankel:
Is the demo link up to date? I ran the code locally and found that the persistent drawers were behaving like temporary drawers 🤔

brad-decker:
@AriLFrankel that'd be because i forgot to update the demo page to use the right props ;P

brad-decker:
https://docs-qrvtwfdpdn.now.sh updated link

AriLFrankel:
LGTM 👍

PR #86, opened 2018-01-05
Add snackbar

kristajg:
Adds the basic Snackbar component, specced for desktop. Includes animations.
Docs:
- open: boolean to open snackbar
- onRequestClose: function that sets open to false
- message: string for snackbar message. can alternatively pass in children to Snackbar
- autoHideDuration: pass in number in milliseconds; default is 4 seconds

Figured I'd open this up for review now since it's pretty much good to go EXCEPT for I'm not really sure what the benefits of the Portal component are right now.  I am currently reading through the code but having a simultaneous review would be helpful.
  
https://docs-mqemnkwgsf.now.sh


brad-decker:
Noice! -- you should investigate whether using the Portal component to move this out of the React DOM structure and close to the root of the html would make it the markup more semantic. Not a requirement for your first PR by any means.

kristajg:
@brad-decker word up, I will deffo check that out

kristajg:
@brad-decker Should we be tagging Daniel for reviews here?

src/components/Snackbar.js
```diff
@@ -0,0 +1,87 @@
...
+  bottom: ${props => (props.open && !props.animateOut ? '0px' : '-48px')};
+`;
+
+class SnackbarComponent extends PureComponent {
+  state = {
+    animateOut: false,
+  };
+
+  componentWillReceiveProps(nextProps) {
+    if (nextProps.open !== this.props.open && nextProps.open === true) {
```
brad-decker: NIT: simplify to (nextProps.open && !this.props.open) -- this will confirm that nextProps.open is true and this.props.open is false.

src/components/Snackbar.js
```diff
@@ -0,0 +1,87 @@
+import React, { PureComponent } from 'react';
+import styled, { keyframes } from 'styled-components';
+// import elevation from '../../mixins/elevation';
+// import { Portal } from '../Portal';
```
martahj: left some commented code in - was this on purpose?

src/components/Snackbar.js
```diff
@@ -0,0 +1,87 @@
...
+`;
+
+class SnackbarComponent extends PureComponent {
+  state = {
+    animateOut: false,
+  };
+
+  componentWillReceiveProps(nextProps) {
+    if (nextProps.open !== this.props.open && nextProps.open === true) {
+      setTimeout(() => {
```
brad-decker: i would get the timer id from the call back and store it on this, then clear the timeout if its somehow opened again prior to the timer going off. 
```js
if (this.delayedCloseTimer) {
  clearTimeout(this.delayedCloseTimer);
}
this.delayedCloseTimer = setTimeout(() => {...}, duration);
```

The setTimeout inside the setTimeout handler is fine as is, doesn't need the reference.

brad-decker:
@kristajg -- would be good to get him to take a look. We need to generate the now build link and attach it so he can QA

brad-decker:
The primary benefit of the portal is that it moves its children into the root of the dom (or at least closer to root). imagine if you are like 10 components deep in a chain of rendering and you want to put the snackbar there because it makes sense there in terms of code (you are managing state, etc), but the snackbar is actually positioned in a way where it doesn't really belong in that section. A portal will put the markup of that component in a higher level in the dom. This allows you to break out of the inheritance of DOM attributes and is semantically correct in html composition.

Its probably not necessary for this iteration tbh. Its better for like dialogs and stuff that have overlays but i think long term we would put all site-wide UI stuff that attaches/affixes to a side of the screen in the Portal.

martahj:
@kristajg could you create a url for this PR so we can see it in action? 😉

kristajg:
@martahj Yep! Should have included right off the bat, just updated https://docs-dwqeelqmbu.now.sh

brad-decker:
@kristajg -- you need to add a transform: translateX(-50%) to get true center alignment. Theoratically the same could be achieved by setting a margin-left equal to a negative value of half of the width of the snackbar -- but -50% with translateX will work regardless of width of snackbar

kristajg:
New: https://docs-mqemnkwgsf.now.sh

pages/snackbar.js
```diff
@@ -0,0 +1,45 @@
...
+      <MaterialThemeProvider theme={{ primary: '#03A9F4' }}>
+        <div className={this.props.className}>
+          <h1>Snackbar</h1>
+          <ul>
+            <li>open - boolean to open snackbar</li>
+            <li>onRequestClose - function that sets open to false</li>
+            <li>message -
+              string for snackbar message. can alternatively pass in children to Snackbar</li>
+            <li>autoHideDuration - pass in number in milliseconds; default is 4 seconds</li>
+          </ul>
```
martahj: I really like this pattern of documenting the props on this page

src/components/Snackbar.js
```diff
@@ -0,0 +1,88 @@
...
+  componentWillReceiveProps(nextProps) {
+    if (nextProps.open && !this.props.open) {
+      if (this.delayedCloseTimer) clearTimeout(this.delayedCloseTimer);
+
+      this.delayedCloseTimer = setTimeout(() => {
+        // start auto-dismissal counter after snackbar is opened
+        this.setState({ animateOut: true });
+        // run onRequestClose, delayed by animation timing
+        setTimeout(() => this.props.onRequestClose(), 300);
+      }, this.props.autoHideDuration || 4000);
```
martahj: Seems like we should set `this.delayedCloseTimer` to null during this callback since line 42 checks to see if it's truthy or not

src/components/Snackbar.js
```diff
@@ -0,0 +1,88 @@
...
+
+  componentWillReceiveProps(nextProps) {
+    if (nextProps.open && !this.props.open) {
+      if (this.delayedCloseTimer) clearTimeout(this.delayedCloseTimer);
+
+      this.delayedCloseTimer = setTimeout(() => {
+        // start auto-dismissal counter after snackbar is opened
+        this.setState({ animateOut: true });
+        // run onRequestClose, delayed by animation timing
+        setTimeout(() => this.props.onRequestClose(), 300);
```
martahj: NIT: could rewrite this line:
```js
setTimeout(this.props.onRequestClose, 300)
```

pages/snackbar.js
kristajg: thanks!

src/components/Snackbar.js
kristajg: I don't understand why?

src/components/Snackbar.js
kristajg: Does this improve performance?

src/components/Snackbar.js
martahj: If we're using the truthiness of `this.delayedCloseTimer` to indicate whether or not there's an active countdown, it might make sense to set it to falsy when the countdown finishes

src/components/Snackbar.js
martahj: Not meaningfully as far as I know, it's just shorter lol. It doesn't create a new function is which is probably good.

src/components/Snackbar.js
kristajg: Hmm. Does it make sense to set something to falsey inside it's own callback..?

kristajg: Unless I'm not 100% following where you'd like to set it to null

martahj: I was thinking something along the lines of:
```js
  componentWillReceiveProps(nextProps) {
    if (nextProps.open && !this.props.open) {
      if (this.delayedCloseTimer) clearTimeout(this.delayedCloseTimer);

      this.delayedCloseTimer = setTimeout(() => {
        // start auto-dismissal counter after snackbar is opened
        this.setState({ animateOut: true });
        // run onRequestClose, delayed by animation timing
        setTimeout(() => this.props.onRequestClose(), 300);
        // reset this.delayedCloseTimer
        this.delayedCloseTimer = null;
      }, this.props.autoHideDuration || 4000);
    } else {
      this.setState({ animateOut: false });
    }
  }
```
Because line 44 will set `this.delayedCloseTimer` to the timer id synchronously at the time when `setTimeout` is called, I'm not aware of any reason that setting `this.delayedCloseTimer` inside the callback would cause a problem.

src/components/Snackbar.js
brad-decker: In isolation its probably not worthwhile improvement in performance, however at scale with multiple places through a codebase doing this it could negatively impact performance (but still small). In this case the cost to refactor is small so i'd vote for going ahead and doing it. In some cases this would be a pre optimization that we wouldn't necessarily enforce. (such as when there were cases in the old app that we were using .bind and not doing that would require a significant rewrite)

src/components/Snackbar.js
brad-decker: The goal is to clear the timeout if it exists at the time the props change in a way that would trigger a new timeout. It would be sufficient to just set the timer variable to null after calling clearTimeout. no need to do it within the timer callback itself imo.

PR #88, opened 2018-01-05
refactor(Avatar): Make avatar more versatile

dekkofilms:
Avatar can now take in a size property, and
refactored the styles to not be absolute positioned.
This will help with it being placed in other components.
Cards Primary component also was affected by this change,
and was fixed accordingly.

Link to preview app: https://docs-tvbesfawbx.now.sh/


pages/cards.js
```diff
@@ -72,21 +64,24 @@ const CardsPage = ({ className }) => (
...
-        </Primary>
-        <SupportingText>Praesent commodo cursus magna, vel scelerisque nisl consectetur et. Integer posuere erat a ante venenatis dapibus posuere velit aliquet. Donec id elit non mi porta gravida at eget metus. Maecenas sed diam eget risus varius blandit sit amet non magna.</SupportingText>
+        <Primary
+          title="Title"
+          subheader="Subhead"
+        />
+        <SupportingText>Praesent commodo cursus magna, vel scelerisque nisl consectetur et.
+        Integer posuere erat a ante venenatis dapibus posuere velit aliquet.
+        Donec id elit non mi porta gravida at eget metus. Maecenas sed diam eget
+        risus varius blandit sit amet non magna.</SupportingText>
```
martahj: NIT: multiline elements should have the children on a different line from the tags:

```js
<SupportingText>
  Praesent commodo cursus magna, vel scelerisque nisl consectetur et.
</SupportingText>
```

src/components/Avatar.js
```diff
@@ -2,23 +2,27 @@ import React from 'react';
 import styled from 'styled-components';
 
 const AvatarComponent = props => (
-  <div className={`${props.className} smc-avatar`}>
+  <div className={`${props.className} smc-avatar`} size={props.size}>
```
martahj: The styled component wraps the AvatarComponent, so you don't need to pass `props.size` in order to be able to access `props.size` in the `styled` call.

src/components/List/ListItem.js
```diff
@@ -2,26 +2,28 @@ import React from 'react';
 import styled from 'styled-components';
 import { Avatar } from '../Avatar';
 
-const ListItemComponent = ({ className, children, leftAvatar }) => (
+const ListItemComponent = ({ className, children, leftAvatar, size }) => (
   <li className={className}>
     {leftAvatar && (
-      <Avatar src={leftAvatar} />
+      <div className="smc-list-avatar">
+        <Avatar src={leftAvatar} size={size} />
```
martahj: Since `size` is not used anywhere else in this component, it might make more sense to leave this out and let the user modify the height and width of the avatar div using `styled()`.

If you leave the prop in, you should rename it to something like `avatarSize` to help the user know what component they are setting the size on.

pages/cards.js
dekkofilms: Good catch ... I think I need to configure something in VSCode cause prettier seems to change my formatting to the file whenever I copy and paste something around.

src/components/Avatar.js
dekkofilms: Ah! Good to know!

brad-decker:
The avatars on the list page are not stretched to fill the circle that contains them. You have gray bars on top and bottom

dekkofilms:
@brad-decker eek! fixed that ... although I don't think that was the greatest picture for an example. I know `object-fit` doesn't have IE or Edge support. I can definitely think through another solution if that isn't appealing.

@martahj fixed based on your comments.

src/components/Card/Primary.js
```diff
@@ -1,13 +1,59 @@
...
-import styled from 'styled-components';
+import styled, { css } from 'styled-components';
+import { Avatar } from '../Avatar';
+import typography from '../../mixins/typography';
 
-const PrimaryComponent = props => (
-  <section className={`${props.className} card-primary`}>{props.children}</section>
+const Title = styled.h1`
+  ${typography('body2')}
+  margin: -.063rem 0;
```
AriLFrankel: Where does this number come from?

src/components/Card/Primary.js
```diff
@@ -1,13 +1,59 @@
...
+
+  ${props => props.large && css`
+    ${typography('headline')}
+    margin: 0;
+  `}
+
+  .primary & {
```
AriLFrankel: I don't think this is taking effect.
 
There are other areas in the Card code where similar selectors are used and don't seem to take effect either. It might be a separate PR to remove those.

dekkofilms:
@brad-decker update link: https://docs-uycvryndav.now.sh/

src/components/Card/Primary.js
```diff
@@ -1,13 +1,59 @@
...
+    ${props => props.large && css `
+      padding-top: 8px;
+    `}
+  }
+`;
+
+const Subheader = styled.h2`
+  ${typography('body1')}
+  margin: -.063rem 0;
```
AriLFrankel: same q as above

src/components/Card/Primary.js
dekkofilms: Yeah, I am not too sure currently ... I can look into it though. I copy and pasted that over from the Title component that isn't needed anymore if this PR goes through.

src/components/Card/Primary.js
dekkofilms: Yeah, good point. I don't mind fixing that stuff up in this PR ... but I was primarily just focused on fixing the styles for the changed Avatar to still work with the Cards.

brad-decker: No need for this pr.

src/components/Card/Primary.js
brad-decker: @AriLFrankel the pr actually didn't touch this stuff, @dekkofilms just messed with how it was composed. If we need to tweak the card component we can do so in another PR

brad-decker:
@dekkofilms rebase, and LGTM

PR #94, opened 2018-01-12
feat(Divider): Add divider

martahj:
Adds the divider component as a standalone component as well as an
option for listitem.

Preview app here: https://docs-irbmjcdpwv.now.sh

### DIFF
--- a/src/components/List/ListItem.js
+++ b/src/components/List/ListItem.js
@@ -1,15 +1,19 @@
 import React from 'react';
 import styled from 'styled-components';
 import { Avatar } from '../Avatar';
+import Divider from '../Divider';
 
-const ListItemComponent = ({ className, children, leftAvatar, avatarSize }) => (
+const ListItemComponent = ({ className, children, leftAvatar, avatarSize, withDivider }) => (
   <li className={className}>
     {leftAvatar && (
       <div className="smc-list-avatar">
         <Avatar src={leftAvatar} size={avatarSize} />
       </div>
     )}
-    <div className="smc-list-content">{children}</div>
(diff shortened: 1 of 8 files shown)


src/components/List/ListItem.js
```diff
@@ -1,15 +1,19 @@
...
   <li className={className}>
     {leftAvatar && (
       <div className="smc-list-avatar">
         <Avatar src={leftAvatar} size={avatarSize} />
       </div>
     )}
-    <div className="smc-list-content">{children}</div>
+    <div className="smc-list-content">
+      {children}
+      {Boolean(withDivider) && <Divider />}
```
dekkofilms: Would we want to be able to pass inset here? I know one use case that people might want is for the divider to also be under the avatar as well. We could just have it as a direct child of the `<li>` and the inset could shift it from under the Avatar, but I have also seen people just use the `<Divider>` component itself after each `<ListItem>` and wouldn't even need it built into the component at all. Thoughts?

martahj: This implementation was based on the material docs, which mention that dividers [should be inset if used with avatars](https://material.io/guidelines/components/dividers.html#dividers-usage). The idea was that it would automatically have the desired behavior (inset if avatars, full-bleed if not) depending on whether the user provided an avatar, and would expand/contract automatically with the list item text if the user set custom width behavior for that.

I could see someone wanting to have more control over that behavior, but my vote would be to move forward with this approach for now if you're comfortable with that.

dekkofilms: Awesome! Sounds good.

PR #95, opened 2018-01-12
feat(Textfield): Add textarea option

martahj:
Adds an option to create a textarea rather than a text field. Updates
a couple textfield styles to facilitate this.

Preview link: https://docs-iqiblnualp.now.sh/


dekkofilms:
Looks good ... I would approve, but it will just become stale. :)

martahj:
@dekkofilms I haven't looked into that at all, no

martahj:
@dekkofilms rebased 👍 I forgot to make a separate commit for it but this PR also fixes the scrolling issue we were seeing

dekkofilms:
Awesome @martahj and I think @ashwoodall was asking about the resize thing. :)

PR #101, opened 2018-01-12
feat(Textfield autoresize): Add multiline autoresize on text area

ashwoodall:
Update text area to autoresize when content is added per material specs. Currently works only with single row. TODO: add resize when rows attribute is specified. 

Preview link: https://docs-vaihimcipu.now.sh/

### DIFF
--- a/src/components/TextField.js
+++ b/src/components/TextField.js
@@ -8,16 +8,21 @@ class TextFieldComponent extends PureComponent {
     focus: false,
     error: this.props.error || false,
     hasBeenFocused: false,
+    height: '100%',
   };
+  
+  textArea = null;
 
   onChange = (e) => {
     this.props.onChange && this.props.onChange(e);
     const text = e.target.value;
     const isInvalid = this.props.validator && !this.props.validator(text);
     const isEmptyButRequired = this.props.required ? !e.target.value : false;
+    const newHeight = this.textArea.scrollHeight;
     this.setState({
       text,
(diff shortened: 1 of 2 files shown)


ashwoodall:
My original pass at this triggered the textarea component in the ternary when the multiline prop was passed down. As it is right now, if no rows are specified it will autoresize, but if rows are passed down it will obey the row count. I can definitely update this to only trigger when multiline is specified though if that's more desirable

src/components/TextField.js
```diff
@@ -8,16 +8,20 @@ class TextFieldComponent extends PureComponent {
...
     hasBeenFocused: false,
+    height: '100%',
   };
 
   onChange = (e) => {
     this.props.onChange && this.props.onChange(e);
     const text = e.target.value;
     const isInvalid = this.props.validator && !this.props.validator(text);
     const isEmptyButRequired = this.props.required ? !e.target.value : false;
+    const newHeight = document.querySelector('.smc-text-field-area').scrollHeight;
```
brad-decker: Might be some danger using this if there are more than one text area on the page. Would be better to get the ref of the element using react ref api and then checking .scrollHeight on the ref onChange. Thoughts?

brad-decker:
@martahj please re-review after rebase

ashwoodall:
Rebased! @martahj please note that now this is optional when "multiline" is passed to textarea

ashwoodall:
@martahj voila! https://docs-vaihimcipu.now.sh/text-fields/

PR #104, opened 2018-01-17
feat(Appbar): Add basic Appbar with title

### LINKED ISSUE #103: Make AppBar Component
https://material.io/guidelines/layout/structure.html#structure-app-bar

ashwoodall:
First pass at Appbar. Currently accepts title props.

TODO:
[] NavIcon
[] Action icons
[] Menu icon

Resolves #103 

Demo: https://docs-iydvzigxai.now.sh

### DIFF
--- a/src/components/Appbar.js
+++ b/src/components/Appbar.js
@@ -1,0 +1,50 @@
+import React, { PureComponent } from 'react';
+import styled from 'styled-components';
+import typography from '../mixins/typography';
+import elevation from '../mixins/elevation';
+
+// placeholder for implementing navicon img
+const NavIcon = styled.div`
+  height: 24px;
+  width: 24px;
+  margin: 12px;
+`;
+
+const TitleComponent = ({ className, children }) => (
+  <h1 className={`${className} smc-appbar-title`}>
+    {children}
+  </h1>
+);
+
+const Title = styled(TitleComponent)`
+  margin: 0 12px;
+   ${typography('title')}
+`;
+
+class AppbarComponent extends PureComponent {
+  render() {
+    return (
+      <div
+        className={`${this.props.className} smc-appbar-container`}
+      >
+        <NavIcon />
+        <Title>{`${this.props.title}`}</Title>
+      </div>
+    );
+  }
+}
+
+const Appbar = styled(AppbarComponent)`
+  top: 0;
+  z-index: 2;
+  display: flex;
+  align-items: center;
+  padding: 0 24px;
+  background-color: ${props => props.theme.primary};
+  ${elevation(2)}
+  height: 64px;
+  position: sticky;
+  color: #fff;
+`;
+
+export default Appbar;
--- a/pages/appbar.js
+++ b/pages/appbar.js
@@ -1,0 +1,55 @@
+import React, { Component } from 'react';
+import styled from 'styled-components';
+import MaterialThemeProvider from '../src/theme/ThemeProvider';
+import { Drawer } from '../src/components/Drawer';
+import Button from '../src/components/Button';
+import Appbar from '../src/components/Appbar';
+
+const PageContainer = styled.div`
+  width: 100vw;
+`;
+
+const Content = styled.div`
+  padding: 10px;
+`;
+
+export default class AppbarPage extends Component {
+  state = {
+    persistentRight: false,
+  };
+
+  togglePersistentRight = () => this.setState(prevState => ({
+    persistentRight: !prevState.persistentRight,
+    persistentLeft: false,
+  }));
+
+
+  render() {
+    return (
+      <MaterialThemeProvider theme={{ primary: '#03A9F4' }}>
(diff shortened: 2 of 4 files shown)


src/components/Appbar.js
```diff
@@ -0,0 +1,46 @@
+import React, { PureComponent } from 'react';
+import styled from 'styled-components';
+import typography from '../mixins/typography';
+
+const AppbarWrapper = styled.div`
+  top: 0;
+  z-index: 2;
+  display: flex;
+  align-items: center;
+  padding: 0 24px 0 24px;
```
AriLFrankel: NIT: this could just be 0 24px;

src/components/Appbar.js
```diff
@@ -0,0 +1,46 @@
...
+
+// placeholder for implementing navicon img
+const NavIcon = styled.div`
+  height: 24px;
+  width: 24px;
+  margin: 12px;
+`;
+
+const Title = styled.h1`
+  margin: 0 12px 0 12px;
```
AriLFrankel: see above

src/components/Appbar.js
```diff
@@ -0,0 +1,46 @@
...
+        <NavIcon />
+        <Title>{`${this.props.title}`}</Title>
+      </AppbarWrapper>
+    );
+  }
+}
+
+const Appbar = styled(AppbarComponent)`
+  background-color: ${props => props.theme.primary};
+  box-shadow: 0 2px 5px rgba(0,0,0,.26);
```
AriLFrankel: This might be best to do with the elevation mixin for consistency.

src/components/Appbar.js
```diff
@@ -0,0 +1,46 @@
...
+const NavIcon = styled.div`
+  height: 24px;
+  width: 24px;
+  margin: 12px;
+`;
+
+const Title = styled.h1`
+  margin: 0 12px 0 12px;
+  ${typography('title')}
+`;
```
martahj: While this pattern works pretty well, one potential issue is that the user of the component isn't given any way to override the styles of components declared using the `styled.h1` syntax because it doesn't add a classname to the component.

One pattern people have been using is to first declare a component with a distinct class name and then wrap that component in a styled-component:
```js
const TitleComponent = ({ className, children }) => (
  <h1 className={`${className} smc-appbar-title`}>
    {children}
  </h1>
);

const Title = styled(TitleComponent)`
  margin: 0 12px 0 12px;
   ${typography('title')}
`
```
This is exactly the pattern that you're using with the `Appbar` itself, but you can use it with the subcomponents to give the user more flexibility over them.

Another pattern is to just render an element with a classname inside the main component, and style it in the main call to `styled()`.

```js
class AppbarComponent extends PureComponent {
  render() {
    return (
      <AppbarWrapper
        className={`${this.props.className} smc-appbar-container`}
      >
        <NavIcon />
        <h1 className="smc-appbar-title">{`${this.props.title}`}</h1>
      </AppbarWrapper>
    );
  }
}

const Appbar = styled(AppbarComponent)`
  margin: 0 12px 0 12px;
  ${typography('title')}

  > .smc-appbar-title {
    margin: 0 12px 0 12px;
    ${typography('title')}
  }
`;
```

src/components/Appbar.js
```diff
@@ -0,0 +1,46 @@
...
+import styled from 'styled-components';
+import typography from '../mixins/typography';
+
+const AppbarWrapper = styled.div`
+  top: 0;
+  z-index: 2;
+  display: flex;
+  align-items: center;
+  padding: 0 24px 0 24px;
+`;
```
martahj: Since you pass a styled-components classname directly into the appbar wrapper component on line 28, you could move all of these styles into the main call to `styled()`.

```js
class AppbarComponent extends PureComponent {
  render() {
    return (
      <div
        className={`${this.props.className} smc-appbar-container`}
      >
        <NavIcon />
        <Title>{`${this.props.title}`}</Title>
      </div>
    );
  }
}

const Appbar = styled(AppbarComponent)`
  background-color: ${props => props.theme.primary};
  box-shadow: 0 2px 5px rgba(0,0,0,.26);
  position: sticky;
  height: 64px;
  color: #fff;
  top: 0;
  z-index: 2;
  display: flex;
  align-items: center;
  padding: 0 24px 0 24px;
`;
```

src/components/Appbar.js
ashwoodall: Good catch! Didn't see that one. Will update

src/components/Appbar.js
ashwoodall: This is really cool and makes sense. I did start to have this thought but didn't do much research in the spirit of just getting done what we needed for MVP. I'm so glad understand these patterns; thanks for the info! I've updated this with the first pattern

ashwoodall:
PR has been updated @AriLFrankel and @martahj - thanks for reviewing!

PR #107, opened 2018-01-23
Begin working on Chips

petegivens:


brad-decker:
@petegivens there's a script in the package.json to publish to now.sh, let me know if you need any assistance!

petegivens:
whoops, sorry, forgot to paste it in.
https://docs-bxbmyltccf.now.sh
I am still working on the delete button--svgs in react are confusing :)

petegivens:
got the Delete Icon working, here's an updated Now link: https://docs-nuflqzbtut.now.sh

brad-decker:
<3 this looks pretty dope. Gonna yell at circle to try rebuilding this project and see if linting and stuff passes. @martahj would you mind doing a code review when you get a chance?

petegivens:
moved the on-click to the label span to avoid invoking both onClick and onDelete. added example of chip with Avatar + delete icon
https://docs-gnnpqjxvtc.now.sh

src/components/Chip.js
```diff
@@ -0,0 +1,96 @@
...
+  height: 32px;
+  color: #616161;
+  background-color: #bdbdbd;
+  margin-right: -4px;
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  border-radius: 50%;
+  margin: 0;
+  padding: 0;
```
AriLFrankel: What is the reason for this? I'm not sure that it has any effect

src/components/Chip.js
```diff
@@ -0,0 +1,96 @@
...
+  ${typography('body')} font-size: 16px;
+`;
+
+const Label = styled.span`
+  font-family: 'Roboto';
```
AriLFrankel: Not related to this PR, but this does highlight the need for fonts in the default theme

src/components/Chip.js
```diff
@@ -0,0 +1,96 @@
...
+  margin-left: 12px;
```
AriLFrankel: NIT: you could consolidate these two margin styles to 0px 12px

src/components/Chip.js
```diff
@@ -0,0 +1,96 @@
+import React, { PureComponent } from 'react';
+import styled from 'styled-components';
+import typography from '../mixins/typography';
+
+const ChipContainer = styled.div`
```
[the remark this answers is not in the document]
petegivens: yes, I agree. I almost added them in but I wasn't sure if they should be added to theme/colors or to the defaultTheme. Also not sure what to name them--is there some standard of color names from Google?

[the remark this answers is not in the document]
AriLFrankel: I would use the colors in the defaultTheme already where possible and where the use case isn't covered either don't worry about it or add them to the theme 👍

ashwoodall:
Agree with @AriLFrankel's comments, but otherwise LGTM

src/components/Chip.js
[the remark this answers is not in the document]
brad-decker: We can also iterate on this in a future PR to add refinements to theming -- we have been attempting to iterate quickly so some theming considerations aren't present in some of the existing components. if you wanna take a stab at this please do but this is mergeable without that and we can add an issue for theming of various components

petegivens:
Implemented all of Ari's comments except for pulling in the colors from the default theme (most of them weren't present). I'm still not really clear whether I should be adding colors directly to the theme and making up my own names or if I should try to pattern match makeMaterialSliderTheme.js?

The self-deleting chips will now set themselves to `display:none` if the user passes in a `removable` prop but does not pass in an `onDelete` prop. 

New Now Link: https://docs-mysccqkets.now.sh

src/components/Chip.js
```diff
@@ -0,0 +1,115 @@
...
+    <path d="M12 2C6.47 2 2 6.47 2 12s4.47 10 10 10 10-4.47 10-10S17.53 2 12 2zm5 13.59L15.59 17 12 13.41 8.41 17 7 15.59 10.59 12 7 8.41 8.41 7 12 10.59 15.59 7 17 8.41 13.41 12 17 15.59z" />
+  </DeleteIconSvg>
+);
+
+class ChipComponent extends PureComponent {
+  state = {
+    removed: false,
+  };
+
+  handleDeleteIconClick = (e) => {
```
AriLFrankel: To me it makes sense that the onDelete is a hook that allows the user to handle the deletion event. I don't see this as mutually exclusive with the UI change happening within the chip.

That means here that it could be

```
if (onDelete) {
  onDelete(e)
}

if(this.props.removable){
  this.setState({removed:true});
}
```

src/components/Chip.js
```diff
@@ -0,0 +1,115 @@
...
+
+const DeleteIcon = ({ onDelete }) => (
+  <DeleteIconSvg onClick={onDelete}>
+    <path d="M12 2C6.47 2 2 6.47 2 12s4.47 10 10 10 10-4.47 10-10S17.53 2 12 2zm5 13.59L15.59 17 12 13.41 8.41 17 7 15.59 10.59 12 7 8.41 8.41 7 12 10.59 15.59 7 17 8.41 13.41 12 17 15.59z" />
+  </DeleteIconSvg>
+);
+
+class ChipComponent extends PureComponent {
+  state = {
+    removed: false,
```
AriLFrankel: In the event that a user does want to control the deletion state outside of the Chip, I think that we need to respect that. This does not allow a user to delete the chip with props.

brad-decker:
@AriLFrankel @petegivens -- we can refine the deletion mechanism after the fact. Lets track an issue to investigate the deletion in real world scenarios and document Ari's change requests for a future PR

brad-decker:
@petegivens could you do me one last favor and delete your yarn lock file and then run yarn again? Something happened with our circle ci and doing this fixed it for another PR

petegivens:
@brad-decker @AriLFrankel I have implemented Ari's delete suggestions. I think his points are valid. My only concern is that the API is getting a little hairy... we have an `onDelete` prop for users to pass in their own action, a `removable` prop which means that the chip will hide itself when clicked, and a `removed` prop to allow users to manage the deletion from the parent component. Let me know what you guys think.

AriLFrankel:
🙌 💯 
This looks great. Pending fixing the yarn.lock I'm 👍 on this.

One other request would be if you could please squash all of your commits together that'd be awesome so we can keep the history pretty clean.

Thanks @petegivens

brad-decker:
One last refinement and then we are merging this. Just rename your commit to this:
```
feat(Chip): implements chip component

Longer description here about the feature, doesn’t have to be intense
```

The above first line of our commit message pattern follows a pattern of type(component/subsystem) to the point msg about commit. The break and then additional details allows historical back tracing through commit history to see the intent of each commit.

In this case yours is a feat or new feature added. 

Thanks again for your contribution looking forward to using it soon!

petegivens:
@martahj i'm not totally sure how to link issues and PRs but this directly addresses https://github.com/ConciergeAuctions/styled-material-components/issues/51

Also, I guess I should have started by reading the issues because I see that you were originally going to call it SingleLineChip. Would you like me to refactor, or should we just leave this as Chip and we can augment it with ContactChip ?

brad-decker:
We;re good to leave it as chip and we can refactor/reword/incorporate it at a later time. merging now :+1:

PR #109, opened 2018-01-24
 feat(Grid List): Add Grid List

ashwoodall:
This PR:

- Moves Husky back where it belongs in devDependencies 
- Updates index to use List and default font to be Roboto
- Adds MVP Grid List implementation 

Check it out: https://docs-ozyvfbfosa.now.sh


pages/lists.js
```diff
@@ -40,31 +40,34 @@ const ListsPage = ({ className }) => (
...
-              </div>
-            } />
+            <ListItemText
+              primary="Primary Title" secondary={
+                <div>
+                  <p>(214) 520-8300</p>
+                  <p>5015 Tracy Street Suite 102</p>
+                  <p>Dallas, Texas 75205</p>
+                </div>
+              } />
```
martahj: Are these just prettier changes?

src/components/GridList/GridTile.js
```diff
@@ -0,0 +1,28 @@
+import React from 'react';
+import styled from 'styled-components';
+
+const GridTileComponent = ({ className, children }) => (
+  <div className={className}>
+    <div className="smc-grid-tile-content">
```
martahj: What problem does the nested div solve? I'm not savvy enough with CSS to tell just by looking 😅

pages/lists.js
dekkofilms: Weird ... it just moved it down a line.

src/components/GridList/GridTile.js
ashwoodall: Ah that was leftover from when I was trying to think about Headers, but then decided it wasn't MVP. I will update 😇

src/components/GridList/GridTile.js
```diff
@@ -0,0 +1,28 @@
...
+  height: '100%',
+  overflow: 'hidden',
+  align-items: center;
+  margin: 0px 2px;
+  min-height:0
+
+  > .smc-grid-tile-content {
+    width: 100%;
+    height: 100%;
+    min-height:0
```
danielstclair: On inspection I was getting an "Improper value" error. I think it should be this instead:
```position: relative;
  display: block;
  height: 100%;
  overflow: hidden;
  align-items: center;
  margin: 0 2px;
  min-height: 0;

  > .smc-grid-tile-content {
    width: 100%;
    height: 100%;
    min-height:0;
  }

File: pages/lists.js (hunk 1 above)
Comment (ashwoodall) [reply pages/lists.js]: yeah

File: src/components/GridList/GridTile.js (hunk 3 above)
Comment (ashwoodall) [reply src/components/GridList/GridTile.js]: updated!

File: src/components/GridList/GridTile.js (hunk 2 above)
Comment (ashwoodall) [reply src/components/GridList/GridTile.js]: updated!

File: src/components/GridList/GridTile.js [hunk 4]
```diff
@@ -0,0 +1,26 @@
...
+import styled from 'styled-components';
+
+const GridTileComponent = ({ className, children }) => (
+  <div className={`${className} smc-grid-tile-content`}>
+    {children}
+  </div>
+);
+
+const GridTile = styled(GridTileComponent)`
+  position: 'relative',
```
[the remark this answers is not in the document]
ashwoodall: dang, sorry I've lived in JSS land for so long! Good catch.

PR #110, opened 2018-01-25
fix(Drawer): Export default

ashwoodall:


martahj:
@ashwoodall Might as well bump the version in this PR too

PR #111, opened 2018-01-26
feat(Checkbox): Add Checkbox

petegivens:
addresses https://github.com/ConciergeAuctions/styled-material-components/issues/32

Now Link: https://docs-lxkvxelsyq.now.sh/

Checkboxes can be controlled or uncontrolled.
Accepts the following props:
- primary: render in primary theme color
- accent: render in accent theme color (default behavior)
- defaultChecked
- checked: for controlled checkboxes
- disabled
- indeterminate
- onChange: for controlled checkboxes
- value

TODO:
 - animations (halp)
 - labels

### DIFF
--- a/pages/checkboxes.js
+++ b/pages/checkboxes.js
@@ -1,0 +1,58 @@
+import React, { PureComponent } from 'react';
+import MaterialThemeProvider from '../src/theme/ThemeProvider';
+import Checkbox from '../src/components/Checkbox';
+import List from '../src/components/List/List';
+import ListItem from '../src/components/List/ListItem';
+import defaultTheme from '../src/theme/defaultTheme';
+
+class CheckboxesPage extends PureComponent {
+  state = {
+    checked: true,
+  };
+
+  handleChange = () => {
+    const checked = !this.state.checked;
+    this.setState({ checked });
+  };
+
+  render() {
+    const { checked } = this.state;
+    return (
+      <MaterialThemeProvider theme={defaultTheme}>
+        <List>
+          <h1 style={{ marginLeft: 25 }}>Checkboxes</h1>
+          <ListItem>
+            <Checkbox id="checkbox2" />
+            <label htmlFor="checkbox2">Accent (default)</label>
+          </ListItem>
+          <ListItem>
+            <Checkbox primary label="Click Me" id="checkbox1" />
+            <label htmlFor="checkbox1">Primary</label>
+          </ListItem>
+          <ListItem>
+            <Checkbox defaultChecked id="checkbox3" />
+            <label htmlFor="checkbox3">Defaults to Checked</label>
+          </ListItem>
+          <ListItem>
+            <Checkbox disabled id="checkbox4" />
+            <label htmlFor="checkbox4">Disabled</label>
+          </ListItem>
+          <ListItem>
+            <Checkbox checked disabled id="checkbox5" />
+            <label htmlFor="checkbox5">Checked and Disabled</label>
+          </ListItem>
+          <ListItem>
+            <Checkbox checked={checked} onChange={this.handleChange} id="checkbox6" />
+            <label htmlFor="checkbox6">Controlled Checkbox</label>
+          </ListItem>
+          <ListItem>
+            <Checkbox indeterminate id="checkbox7" />
+            <label htmlFor="checkbox7">Indeterminate Checkbox</label>
+          </ListItem>
+        </List>
+      </MaterialThemeProvider>
+    );
+  }
+}
+
+export default CheckboxesPage;
--- a/src/components/Checkbox.js
+++ b/src/components/Checkbox.js
@@ -1,0 +1,142 @@
+import React, { PureComponent } from 'react';
+import styled from 'styled-components';
+
+class Checkbox extends PureComponent {
+  state = {
+    checked: this.props.checked || this.props.defaultChecked || false,
+    indeterminate: this.props.indeterminate || false,
+  };
+
+  handleInputChange = (e) => {
+    const checked = e.target.checked;
+    this.setState({ indeterminate: false, checked });
+
+    if (this.props.onChange) {
+      this.props.onChange(e);
+    }
+  };
+
+  render() {
+    const { primary, disabled, checked: checkedProp, value, id } = this.props;
+    const { indeterminate } = this.state;
+    // determine if checkbox is controlled or uncontrolled
+    const checked = checkedProp !== undefined ? checkedProp : this.state.checked;
+
+    return (
+      <CheckboxContainer
+        className="smc-checkbox-container"
+        primary={primary}
+        disabled={disabled}
+        checked={checked}
+      >
+        <CheckboxBackground
+          className="smc-checkbox-background"
+          primary={primary}
+          checked={checked}
+          disabled={disabled}
+          indeterminate={indeterminate}
+        >
+          {indeterminate && <IndeterminateMark className="smc-checkbox-indeterminate-mark" />}
+          {checked && !indeterminate && <CheckMark className="smc-checkbox-check-mark" />}
+        </CheckboxBackground>
+        <Input
+          onChange={this.handleInputChange}
+          disabled={disabled}
+          checked={checked}
+          value={value}
+          id={id}
+        />
+      </CheckboxContainer>
+    );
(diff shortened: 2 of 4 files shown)


petegivens:
Hopefully this is a decent start on checkboxes. I haven't done anything with styling the labels yet. Not sure what approach you guys think we should take on that. 

I think it would be cool to bundle them and have the checkbox accept the label as a prop.  I personally find <labels> annoying to work with, so I would enjoy it.  We could accept a prop to specify whether the label should display before or after the checkbox, but it would take away the flexibility to put the label wherever you want. LMK what you guys think.

src/components/Checkbox.js
```diff
@@ -0,0 +1,161 @@
...
+        >
+          {indeterminate && <IndeterminateMark className="smc-checkbox-indeterminate-mark" />}
+          {checked && !indeterminate && <CheckMark className="smc-checkbox-check-mark" />}
+        </CheckboxBackground>
+        <Input
+          onChange={this.handleInputChange}
+          disabled={disabled}
+          checked={checked}
+          value={value}
+        />
```
martahj: Might be good to give a classname to Input as well so the user can style it

martahj:
This looks great.

I don't see a strong need to export a label component at this point. The [material design spec](https://material.io/guidelines/components/selection-controls.html#selection-controls-checkbox) doesn't mention any particular styling of one, so imo it would be a little misleading to export one with baked-in styles. Open to alternative suggestions, though.

src/components/Checkbox.js
petegivens: I actually intentionally didn't expose the classname for the input because it should always stay hidden. Any styling the user needs to do should be applied to one of the visible layers that I threw on top of the input. Basically only using it to allow for uncontrolled usage.

martahj: That makes sense as an approach 👍  However, the input's position, width, and height are hardcoded. If the user can't directly style these, it would be helpful if the input automatically adjusted to be centered in the parent and take up the space provided by the parent so that the clickable area remains where the user expects it to be.

petegivens: ah ok, good call. might need a little help with this one. The checkbox container is 18x18 with a padding of 11px. I can make the input inherit the size, but padding does not seem to be inheritable. Is there any way in S-C or SCSS to tell a class to use the same values as a prop from a different class? If so, I could just lock the size and padding to the parent container. Open to other suggestions if that doesn't work :)

martahj: Hmm.. it looks like padding can take the value [inherit](https://developer.mozilla.org/en-US/docs/Web/CSS/padding). Might be worth a shot.

Another option might be to give both the parent and the input a new classname, and specify the padding and size on that classname. That way, if someone tried to override it, it would impact both. However, it would still be possible for users to not realize this and give the size and padding to only the parent.

If the padding is being tricky, another alternative would be to not have any padding on the parent, but instead to give the checkbox outline a margin... another possibility is to not hardcode the space between the container and the checkbox outline, but instead have it expand propertionally with the size of the component.

Let me know if any of that works - happy to talk through some more options if not :)

brad-decker:
we could have the label prop be able to accept a component in the future, instead of just a text string, giving us the ability to compose it? we could even go so far as to intercept the component and add a classname to it so we can style it similarly to our other nested components.not necessary for this PR.

petegivens:
I think this should address @martahj 's concern. Since we are not exposing the input itself, it must automatically resize so that the clickable area of the checkbox is as big as the user makes the outer container. 

What I've been trying (but so far unable) to find is a way to increase the input's clickable area without increasing its size. Increasing its size makes adjacent text look bad, because they don't line up vertically. 

Very open to suggestions on how to solve this, but maybe this is merge-able now. LMK.

pages/checkboxes.js
```diff
@@ -0,0 +1,51 @@
...
+    const { checked } = this.state;
+
+    return (
+      <MaterialThemeProvider theme={defaultTheme}>
+        <div>
+          <h1>Checkboxes</h1>
+          <form>
+            <Checkbox primary label="Click Me" name="checkbox1" />
+            <label htmlFor="checkbox1">Primary</label>
+            <br />
```
ashwoodall: I prefer wrapping in block-level elements or the List component to `<br /> ` tags, but that's totally just me and not necessary to change here

src/components/Checkbox.js
ashwoodall: I agree with trying to pass a new classname to both elements to try to address this. However, It does seem like the [demo on material.io](https://material-components-web.appspot.com/checkbox.html) is also clickable outside of the box area. Unless I'm misunderstanding the issue here

martahj: @ashwoodall The issue isn't whether or not it's clickable outside the box area, but making sure the clickable area expands and contracts as the box size changes, as well as moves horizontally or vertically as necessary.

ashwoodall: Then yes, I agree with the shared classname approach 👍

src/components/Checkbox.js
```diff
@@ -0,0 +1,163 @@
...
+    }
+  };
+
+  render() {
+    const { primary, accent: accentProp, disabled, checked: checkedProp, value } = this.props;
+    const { indeterminate } = this.state;
+    // determine if checkbox is controlled or uncontrolled
+    const checked = checkedProp !== undefined ? checkedProp : this.state.checked;
+    // if no primary or accent prop is passed in, default to accent
+    const accent = !primary && (accentProp || true);
```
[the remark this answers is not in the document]
petegivens: Good question. I added that after realizing that the checkbox should default to the theme's accent color. I suppose there is really no need for an accent prop at all given that it's the default option.

src/components/Checkbox.js
petegivens: I worry about using a shared classname because of the reason Martah pointed out. The user would have to know to apply the styles to the exposed classes and the "extra" classname we provide. I think the solution is in this technique: https://www.styled-components.com/docs/advanced#referring-to-other-components, but I haven't had much time to experiment. Hopefully we can iterate on this once we figure it out.

In the meantime, let me know what you want to do with the primary/accent prop to merge this one. I suppose it might make more sense to have a `color` prop rather than `primary` and `accent`. Acceptable values could be 'primary', 'accent'(default), or an RGB or hex value. Then we could apply the color to the border outline, checkbox, on-hover, etc. Does that sound like a better approach? @martahj @ashwoodall @brad-decker ?

martahj: Hm.. I see the advantage of that, particularly as an initial solution, but I think there's a danger in moving towards an api where the user specifies styles by passing them in manually instead of going through the theme.

One possible approach would be to keep `primary` and `accept` passed in the way they are, and then hardcode a default fallback color in the component based on the material design example. Alternatively, you could add the fallback color to theme as `checkboxColor` or something similar. It wouldn't be the first component to have theming specific to it (slider does as well). 

Thoughts? @petegivens @ashwoodall @brad-decker

petegivens: That reminds me of a question I was wondering when writing this. Can we assume that a theme will always be present? Even if one ships with the library, wouldn't users be free to use components without wrapping them in the `ThemeProvider`? If users can do that, hard-coding the fallback values per the default material light theme seems like the thing to do. If we can guarantee the theme will always be there, then I would lean towards putting it in the defaultTheme. But, you tell me what works best! 😛

martahj: Yes, we can assume that anyone using this library will have to wrap the components in the `ThemeProvider`. Ultimately we'll want the library to come with things like the material ripple and elevation effects "baked in" (we have them implemented in some components but not others at this point), and those will require a theme to use :)

petegivens: OK thanks! So my personal preference would be to add the colors to the defaultTheme. I'll call it checkbox-something, but it's likely that we'll find other components that use it and maybe it can get a more generic name in the future? I'll submit those changes today, let me know if you want me to go in a different direction.

petegivens:
OK, all colors are pulling from the Theme now. I added a color to our defaultTheme called `disabledCheckbox`.

Also addressed @martahj's point about making the checkmark auto-resize.

Removed all use of the `accent` prop since that's the default behavior and simplified the code a bit.

PR #112, opened 2018-01-29
fix(TextField): Fix scrollHeight error

ashwoodall:
Fix 'Cannot read property 'scrollHeight' of undefined' error received
after compilation

### DIFF
--- a/src/components/TextField.js
+++ b/src/components/TextField.js
@@ -18,7 +18,8 @@ class TextFieldComponent extends PureComponent {
     const text = e.target.value;
     const isInvalid = this.props.validator && !this.props.validator(text);
     const isEmptyButRequired = this.props.required ? !e.target.value : false;
-    const newHeight = this.textArea.scrollHeight;
+    const newHeight = this.props.textarea ? this.textArea.scrollHeight : '';
+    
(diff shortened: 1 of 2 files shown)


src/components/TextField.js
```diff
@@ -18,11 +18,12 @@ class TextFieldComponent extends PureComponent {
...
     const isInvalid = this.props.validator && !this.props.validator(text);
     const isEmptyButRequired = this.props.required ? !e.target.value : false;
-    const newHeight = this.textArea.scrollHeight;
+    const newHeight = this.props.textarea ? this.textArea.scrollHeight : '';
+    
     this.setState({
       text,
       error: this.props.error || isInvalid || isEmptyButRequired,
-      height: !this.props.multiline || this.textArea.value === '' ? '100%' : newHeight,
+      height: !this.props.textarea || this.textArea.value === '' ? '100%' : newHeight,
```
martahj: What's the reason for no longer checking for `this.props.multiline`?

ashwoodall: penguin

brad-decker: I love a good gif comment.

PR #116, opened 2018-02-03
feat(AUTHORS): scripted AUTHORS

AriLFrankel:
This introduces a postcommit hook that will automatically update the AUTHORS file.

This will:
- search the AUTHORS file for the email of the committer
- add a line with name <email> for that committer if they are not represented already

This also introduces a `seed-authors` script that will
- output authors by order of first contribution to the AUTHORS file
- filter out repeat entries (unique by email) and commits accidentally added before being associated with a git account

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -1,14 +1,12 @@
-#Authors
+# Authors orderd by date of first contribution 
 
-Brad Decker <brad.decker@merlinlabs.com>
-David Lee <david.lee@merlinlabs.com>
-Natalia Bernard <natalia.bernard@merlinlabs.com>
-Marta Johnson <marta.johnson@merlinlabs.com>
-Nathan Schwartz <nathan.schwartz@merlinlabs.com>
-Krista Goralczyk <krista.goralczyk@gmail.com>
-Daniel St.Clair <daniel.stclair@merlinlabs.com>
-Jean Page <jean.page@merlinlabs.com>
-Bennett Staley <bennett.staley@merlinlabs.com>
-Ari Frankel <ari.frankel@merlinlabs.com>
-Pete Givens <pgivens@gmail.com>
-Ashley Woodall <ashley.woodall@merlinlabs.com>
+
+ Brad Decker <bhdecker84@gmail.com>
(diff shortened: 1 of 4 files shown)


scripts/seed-authors.sh
```diff
@@ -0,0 +1,22 @@
...
+  echo -e $"# Authors orderd by date of first contribution \n\n" > $root/AUTHORS
+
+  # get authors name <email> contribution date in order
+  # remove bracketed text such as '[user]'
+  # and contributions not associated with a git account (ending in .local)
+  # filter unique by email address
+  # sort by date of contribution again after deduping
+  # remove date of contribution
+  # output to AUTHORS file
+  git log --format='%ad %an <%ae>' --date="short" --reverse | sed -e 's/\[.*\]//' -e '/.local>$/c\' | rev | sort -k 1,1 -u | rev | sort -n -t"-" -k1 -k2 -k3 | sed -e 's/^[0-9\-]*//' >> $root/AUTHORS
```
brad-decker: Dayam... Nice :+1:

AUTHORS
```diff
@@ -1,14 +1,12 @@
...
-Pete Givens <pgivens@gmail.com>
-Ashley Woodall <ashley.woodall@merlinlabs.com>
\ No newline at end of file
+
+ Brad Decker <bhdecker84@gmail.com>
+ iamdavid0091 <iamdavid0091@gmail.com>
+ Cecilia Cisneros <cmvrie@utexas.edu>
+ martahj <marta.hourigan.johnson@gmail.com>
+ Ari Frankel <ari.l.frankel@gmail.com>
+ kristajg <krista.goralczyk@conciergeauctions.com>
```
martahj: Looking at the script to update authors, we should be able to change Krista's back to pointing at `krista.goralczyk@gmail.com` by just manually changing it, and the script won't overwrite it in the future.. is that correct?

AriLFrankel:
@martahj changed Krista's email back to her requested email, and that's correct that it should not be overwritten by a call tp update-authors.

PR #130, opened 2018-02-12
chore(update patterns): Remove classNames

brad-decker:
Goal: Remove the need to create our own components simply to attach a
class name, or other identifiable attribute.

Secondary Goal: Remove classnames, which can cause styling issues, and
instead use data-smc attribute to identify the component name. Where
necessary we should use the component itself in interpolation to
actually style sub components.

This just removes the custom button component. Wrapping a component in
'styled' removes the automatic whitelisting of props that react does.
This removes the need to declare an intermediate click action and white
list the props on the button.

### EDIT:
This is an evolving PR that will contain a lot of API updates. Each API update will be detailed here:

#### Avatar
I decided to decompose this component into one that uses other basic building block components (Circular and Image) to compose itself. This allows for using the reverse selector pattern to keep styles at the level that they should be. It also allows the end user to import Circular and Image from the library and use them for styles. Suggested API:
```js
import { Avatar, Image } from 'styled-material-components';

const CustomAvatar = Avatar.extend`
  width: 400px;
  & ${Image} {
    width: 400px;
  }
`;
```

### DIFF
--- a/src/components/Avatar.js
+++ b/src/components/Avatar.js
@@ -1,29 +1,10 @@
-import React from 'react';
-import styled from 'styled-components';
+import Circular from './Circular';
+import Image from './Image';
 
-const AvatarComponent = props => (
-  <div className={`${props.className} smc-avatar`}>
-    {props.src ? (
-      <img src={props.src} className="smc-avatar-img" />
-    ) : props.children}
-  </div>
-);
-
-export const Avatar = styled(AvatarComponent)`
-  position: relative;
-  display: flex;
-  align-items: center;
-  justify-content: center;
-  width: ${props => props.size || 40}px;
-  height: ${props => props.size || 40}px;
-  background: #bdbdbd;
-  border-radius: 50%;
-  overflow: hidden;
-
-  & .smc-avatar-img {
-    max-width: 100%;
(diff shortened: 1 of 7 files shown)


petegivens:
How would this work for a more complex component? If I were to refactor Checkbox using this patter, should I be attaching the `data-smc` attribute to the outermost container only?

brad-decker:
Gonna stop makign this so huge and do micro PRs for every component as i get to it -- This way @petegivens and others can implement the same patterns on other components as well (I mention @petegivens cause he's already started on the chip)

brad-decker:
@ashwoodall @martahj this is ready for review and merge if all agree and approve.

src/components/Avatar.js
```diff
@@ -1,29 +1,10 @@
...
-  overflow: hidden;
-
-  & .smc-avatar-img {
-    max-width: 100%;
-    width: 100%;
-    height: 100%;
-    object-fit: cover;
-  }
+export const Avatar = Circular.extend.attrs({
+  children: props => (props.src ? <Image src={props.src} /> : props.children),
```
martahj: ✨

PR #131, opened 2018-02-12
docs(Contributing Guide): Adjust grammar

calebjanderson:
Small change but it should be `Anyone` instead of `Any one` in this particular context :)


AriLFrankel:
@calebjanderson thanks for the contribution! 

One request: please install the dependencies and recommit (you could just `commit --amend --no-edit`) so that Husky runs the post-commit hook your name gets its rightful place in the AUTHORS file :)

PR #137, opened 2018-02-13
Fix(button): fix button styles

ceciliamvrie:
Modified: 
* Button

Fixed button text color to not match the background color
of the raised and flat buttons. 

Issues: none

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -12,3 +12,4 @@
  Pete Givens <pgivens@gmail.com>
  Julianne Johnson <tweetypie417@yahoo.com>
  calebjanderson <caleb.anderson130@gmail.com>
+ ceciliamvrie <ceciliamvrie@gmail.com>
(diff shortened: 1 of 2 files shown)


AUTHORS
```diff
@@ -12,3 +12,4 @@
  Pete Givens <pgivens@gmail.com>
  Julianne Johnson <tweetypie417@yahoo.com>
  calebjanderson <caleb.anderson130@gmail.com>
+ ceciliamvrie <ceciliamvrie@gmail.com>
```
martahj: Is your UT email still valid? If not, you might want to remove it from the authorslist

martahj:
Looks good! Could you do the following:
* generate a link to a Now app (you can do this by running `yarn now-build` from the root of the smc directory) and post the link so we can check out your work?
* Bump the version by running `npm version patch` from the same directory

Thanks!

ceciliamvrie:
https://docs-plkeldbnkl.now.sh/chips/ here's the link!

AUTHORS
ceciliamvrie: Ill make sure if it's still valid

ceciliamvrie:
https://docs-fmvozljlng.now.sh

PR #139, opened 2018-02-14
Refactor(chips): Add styles to chips

ceciliamvrie:
Modified: 
* Chips 

Added styles to chips that match more with the spec.
Replaced 'removable' property with 'deletable' to match the spec. 
Also fixed some style warnings with eslint for the Chip component. 

Issues: none

### DIFF
--- a/src/components/Chip.js
+++ b/src/components/Chip.js
@@ -1,6 +1,6 @@
 import React, { PureComponent } from 'react';
 import styled from 'styled-components';
-import typography from '../mixins/typography';
+import Avatar from './Avatar';
 
 const ChipContainer = styled.div`
   display: inline-flex;
@@ -11,6 +11,7 @@ const ChipContainer = styled.div`
   height: 32px;
   border-radius: 16px;
   font-size: 13px;
+  font-weight: 200;
   color: ${props => props.theme.textColors.primary};
   ${props =>
     props.clickable &&
@@ -18,25 +19,16 @@ const ChipContainer = styled.div`
     :hover {
       background-color: #CECECE;
     }`}
+  ${props => props.deletable &&
+  `:hover {
+    background-color: grey;
+    color: white;
+  }`}
   ${props =>
     props.removed &&
     `
     display: none;
   `}
-`;
-
-const Avatar = styled.div`
-  width: 32px;
-  height: 32px;
-  color: #616161;
-  background-color: #bdbdbd;
-  margin-right: -4px;
-  display: flex;
-  justify-content: center;
-  align-items: center;
-  border-radius: 50%;
-  ${typography('body')}
-  font-size: 16px;
 `;
 
 const Label = styled.span`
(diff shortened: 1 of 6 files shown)


ceciliamvrie:
Heres my link! https://docs-plkeldbnkl.now.sh/chips/

ceciliamvrie:
Here is the new link! https://docs-fmvozljlng.now.sh/chips/

src/components/Chip.js
```diff
@@ -52,62 +62,76 @@ const DeleteIconSvg = styled.svg`
...
-    const { avatar, label, onClick, removable, onDelete, removed: removedProp } = this.props;
+    const { avatar, label, onClick, deletable, onDelete, removed: removedProp } = this.props;
     // determine if chip is managing its own removal (uncontrolled)
     // or if it's being managed by the parent (controlled)
     const removed = removedProp !== undefined ? removedProp : this.state.removed;
 
     return (
       <ChipContainer
+        onMouseEnter={this.handleChipHover}
+        onMouseLeave={this.handleChipHover}
```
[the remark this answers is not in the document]
ceciliamvrie: Yeah, that sounds good! Ill change it!

src/components/Chip.js
```diff
@@ -52,62 +62,76 @@ const DeleteIconSvg = styled.svg`
...
+        deletable={deletable}
         className="smc-chip-container"
         onKeyDown={this.handleKeyDown}
         tabIndex={0}
         clickable={onClick}
         removed={removed}
       >
-        {avatar && <Avatar className="smc-chip-avatar">{avatar}</Avatar>}
+        {avatar && <Avatar className="smc-chip-avatar" deletable={deletable} hovered={this.state.hovered}>{avatar}</Avatar>}
```
dekkofilms: [Q] I'm also curious as to why Avatar has deletable is here? I wouldn't think it would matter to the hover whether the chip was deletable or not. Also, wondering if it would be better to target Avatar as a sub-component here // if the Avatar component already built could fit in here instead of defining a new one?

ceciliamvrie: Maybe I should use another prop name instead of 'deletable' for the Avatar component because it implies that it can be deleted, but I'm using it so it's color matches the background of the deletable chip. If not, then it would just stay a lighter grey.

I also like the thought of targeting Avatar as a subcomponent - let me see what I can do!

ceciliamvrie:
Awesome! I just pushed in the changes wanted; here is the link => 
https://docs-mgqxwsapvo.now.sh

petegivens:
Looks great!

I think we can eliminate the onMouseEnter, onMouseLeave, handlers and hover state by using the child selector pattern. 

Inside the DeleteIcon component, give it: 
```
${ChipContainer}:hover & {
    fill: white;
  }
```

Also, I'm not sure that the label should be font-weight 200. The spec lists the label text as `13sp Roboto Regular`

PR #142, opened 2018-02-15
refactor card styles

ceciliamvrie:
Modified: 
• Card

I added the "hover" styles to match the spec whenever 
someone hovers their mouse over the card component.
I made sure that it only effects the cards with the hover 
property in them. 

Issues: none

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "styled-material-components",
-  "version": "0.0.33",
+  "version": "0.0.36",
   "description": "coming soon",
   "main": "dist/index.js",
   "scripts": {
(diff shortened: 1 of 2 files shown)


ceciliamvrie:
Here's the link! => https://docs-plvabcvcyr.now.sh

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "styled-material-components",
-  "version": "0.0.33",
+  "version": "0.0.36",
```
ashwoodall: Any reason for not bumping to 0.0.34?

ceciliamvrie: Hmm I believe that the latest version is .35, and so I was bumping it to .36

martahj: Looks like this [other open PR](https://github.com/ConciergeAuctions/styled-material-components/pull/139/files) bumps the package to .35 (so will need to be merged before this one).. don't know what happened to .34 though 🤔

brad-decker:
:+1: 💯

PR #143, opened 2018-02-15
feat(BottomSheet): Add BottomSheet

haileybee1231:
[View Here](https://docs-exsmtmyrql.now.sh)

Bottomsheets can be either modal or persistent.
To specify a modal sheet, pass a "modal" prop.

TODO:
Add list style bottomsheet
Add list style bottomsheet with header
Add grid style bottomsheet
Add grid style bottomsheet with header
Make persistent sheet scrollable container for content
Implement mobile responsive heights and widths

### DIFF
--- a/pages/bottomsheets.js
+++ b/pages/bottomsheets.js
@@ -1,0 +1,53 @@
+import React, { Component } from 'react';
+import styled from 'styled-components';
+import MaterialThemeProvider from '../src/theme/ThemeProvider';
+import { BottomSheet } from '../src/components/BottomSheet';
+import Button from '../src/components/Button';
+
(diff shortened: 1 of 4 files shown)


pages/bottomsheets.js
```diff
@@ -0,0 +1,53 @@
...
+  padding: 10px;
+`;
+
+export default class PortalPage extends Component {
+  state = {
+    modal: false,
+    persistent: false
+  };
+
+  toggleModal = () => this.setState(prevState => ({
```
brad-decker: Beautiful use of setState with a function to make sure that multiple calls of the function operate on the correct version of state. :+1:

PR #145, opened 2018-02-17
Switch feature and Avatar.js bug fix

### LINKED ISSUE #34: Create Switch component
https://material.io/guidelines/components/selection-controls.html#

jastra4:
https://docs-amnypfkmsw.now.sh/
fixes #34

### DIFF
--- a/src/components/Switch/Switch.js
+++ b/src/components/Switch/Switch.js
@@ -6,15 +6,13 @@ import SwitchThumb from './SwitchThumb';
 class SwitchComponent extends PureComponent {
   state = {
     on: true,
-    off: false,
   };
 
   handleOnClick = () => {
     if (!this.props.disabled) {
-      this.setState({
-        on: !this.state.on,
-        off: !this.state.off,
-      });
+      this.setState(prevState => ({
+        on: !prevState.on,
+      }));
       this.props.onChange && this.props.onChange();
     }
   }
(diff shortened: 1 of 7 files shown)


src/components/Switch/Switch.js
```diff
@@ -0,0 +1,40 @@
...
+
+class SwitchComponent extends PureComponent {
+  state = {
+    on: true,
+    off: false,
+  };
+
+  handleOnClick = () => {
+    if (!this.props.disabled) {
+      this.setState({
```
dekkofilms: Changing state functionally might be the better approach here since we are updating the component’s state using the current state. Using a function makes it more reliable, so for example:

```
this.setState(prevState => ({
    on: !prevState.on
    ...
}));
```

src/components/Switch/Switch.js
```diff
@@ -0,0 +1,40 @@
+import React, { PureComponent } from 'react';
+import styled from 'styled-components';
+import SwitchTrack from './SwitchTrack';
+import SwitchThumb from './SwitchThumb';
+
+class SwitchComponent extends PureComponent {
+  state = {
+    on: true,
```
dekkofilms: Since off is always the opposite of on, we could just eliminate the need for even keeping track of off. We could just keep track of on.

brad-decker:
Very nice work @jastra4

src/components/Switch/Switch.js
jastra4: Thanks Taylor, I appreciate the feedback. I'll make sure to make those changes.

martahj:
@dekkofilms Could you re-review this PR when you get a chance?

Issue #147, opened 2018-02-19
Sidebar Does Not Match Spec


BennettStaley:
https://material.io/components/web/catalog/drawers/ 
https://material-components-web.appspot.com/drawer/persistent-drawer.html

The default size here is 240px.
The default behavior is to always push page content. (when persistent)

Currently our size changes based on media queries, and stops pushing content as the Before and After after styles have their display set incorrectly.

BennettStaley:
Fixed!

PR #148, opened 2018-02-19
Sidebar Spec Fix

BennettStaley:
Adjusts the sidebar spec to reflect the spec defined within material.

Changes:
Width is now fixed to 240px, as spec dictates. Also changes the display to inline-table for the Before/After left and right shifts, meaning the (persistent) sidebar will always take up the space it requires, as spec dictates.

https://docs-hftsbpxxmj.now.sh !

Fixes Issue #147

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -14,3 +14,4 @@
  calebjanderson <caleb.anderson130@gmail.com>
  ceciliamvrie <ceciliamvrie@gmail.com>
  haileybee1231 <haileybee1231@gmail.com>
+ bennett staley <me@bennettstaley.com>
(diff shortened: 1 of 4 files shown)


AUTHORS
```diff
@@ -14,3 +14,4 @@
  calebjanderson <caleb.anderson130@gmail.com>
  ceciliamvrie <ceciliamvrie@gmail.com>
  haileybee1231 <haileybee1231@gmail.com>
+ bennett staley <bennett.staley@conciergeauctions.com>
```
ashwoodall: is this the email you want to list here, @BennettStaley?

Issue #41, opened 2017-11-07
Make ListSubheader Component


martahj:
https://material.io/guidelines/components/subheaders.html#

PR #151, opened 2018-02-20
refactored drawer styles to match spec

ceciliamvrie:
I refactored the styles of the AppBar Drawer component to further match the look of the spec. 
To accomodate for there being two different types of AppBars, I passed a drawer prop through to specify if the AppBar should be a Drawer or not. 

Modified: 
* AppBar 
* Drawer 

Link => https://docs-frjntmeopo.now.sh!

### DIFF
--- a/src/components/Appbar.js
+++ b/src/components/Appbar.js
@@ -10,7 +10,7 @@ const NavIcon = styled.div`
   margin: 12px;
 `;
 
-const TitleComponent = ({ className, children }) => (
+const TitleComponent = ({ className, children, drawer }) => (
   <h1 className={`${className} smc-appbar-title`}>
     {children}
   </h1>
@@ -18,7 +18,12 @@ const TitleComponent = ({ className, children }) => (
 
 const Title = styled(TitleComponent)`
   margin: 0 12px;
-   ${typography('title')}
+  ${typography('title')}
+  ${({ drawer }) => drawer !== undefined && `
+    font-size: 17px;
+    position: absolute;
+    right: 3px;
+  `}
 `;
 
 class AppbarComponent extends PureComponent {
@@ -28,7 +33,7 @@ class AppbarComponent extends PureComponent {
         className={`${this.props.className} smc-appbar-container`}
       >
         <NavIcon />
-        <Title>{`${this.props.title}`}</Title>
+        <Title drawer={this.props.drawer} >{`${this.props.title}`}</Title>
       </div>
     );
   }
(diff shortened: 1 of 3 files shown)


src/components/Drawer.js
```diff
@@ -5,7 +5,7 @@ import elevation from '../mixins/elevation';
 import { Portal } from '../components/Portal';
 
 const drawerSizing = css`
-  width: 240px;
+  width: 320px;
```
ashwoodall: @BennettStaley just updated this component in this PR  -
 https://github.com/MerlinLabs/styled-material-components/commit/eaa73760c359d948219638701739bac14b499672

Does this update it further?

src/components/Appbar.js
```diff
@@ -45,6 +50,11 @@ const Appbar = styled(AppbarComponent)`
   height: 64px;
   position: sticky;
   color: #fff;
+  ${({ drawer }) => drawer && `
+    align-items: flex-end;
+    padding: 12px;
+    height: 155px;
```
ashwoodall: Nice work, but I think this might actually be a subheader component? - https://github.com/MerlinLabs/styled-material-components/issues/41

I know for our team we're using the regular appbar height in our sidebar drawer. Let me know what you think.

src/components/Drawer.js
ceciliamvrie: I was taking a look at the Material Component library examples here, at the drawer component. Feel free to let me know if that is what you guys were shooting towards!

https://material-components-web.appspot.com/drawer/temporary-drawer.html

ceciliamvrie: But yeah, that's why I modified the width of the drawer.

src/components/Appbar.js
ceciliamvrie: Oh, I see. Yeah, so should I change the title of this PR? I didn't realize that I was working on two things at once lol!

ashwoodall: We can definitely talk about it, but I personally think it might be more easily found if it was a separate subheader component.

src/components/Drawer.js
BennettStaley: What the heck, i went off other designs...

https://material.io/components/web/catalog/drawers/
https://material-components-web.appspot.com/drawer/persistent-drawer.html

which are best?

ceciliamvrie: Yeah, I guess I can specify that I am building a temporary drawer, and then make a permanent drawer to keep the older version intact!

ceciliamvrie: Okay, I'm going to go ahead and redo this PR

PR #152, opened 2018-02-21
add NavIcon support for Appbar

BennettStaley:
Adds a navIcon prop to optionally pass in for added functionality determined by the user! (else renders the placeholder as before)

https://docs-kujyejzdul.now.sh


BennettStaley:
not really, its up to the user to pass in a working icon. What is being passed into the sidebar is just an A tag with an 'x' inside it.

martahj:
@BennettStaley Ok then LGTM!

PR #153, opened 2018-02-22
Add temporary drawer

ceciliamvrie:
Modified: 
* Drawer

I re-did the original PR - this one doesn't modify the persistent drawer,
but adds support and styles for a temporary drawer. Thanks @martahj for showing
the correct implementation of how different drawers should be like!

link => https://docs-cdikxdudmw.now.sh!

### DIFF
--- a/pages/appbar.js
+++ b/pages/appbar.js
@@ -56,7 +56,7 @@ export default class AppbarPage extends Component {
               attachment='right'
               open={this.state.temporaryRight}
               handleRequestClose={this.toggleTemporaryRight}>
-              <Appbar title='I am a temporary appbar!' />
+              <Appbar title='I am in this drawer!' />
               <p>I am a <b>temporary</b> drawer!</p>
             </Drawer>
             <Button
--- a/pages/appbar.js
+++ b/pages/appbar.js
@@ -64,7 +64,7 @@ export default class AppbarPage extends Component {
               primary
               onClick={this.togglePersistentRight}
             >
-              Launch Appbar Drawer
+              Launch Permanent Drawer
             </Button>
             <Button
               raised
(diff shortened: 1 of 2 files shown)


pages/appbar.js
```diff
@@ -52,6 +66,13 @@ export default class AppbarPage extends Component {
             >
               Launch Appbar Drawer
```
ashwoodall: If we're adding different types of drawers to the appbar page, should we make a clear difference between the button labels?

ceciliamvrie: Yeah! For the permanent drawer button I can put 'Launch Permanent Drawer'

pages/appbar.js
```diff
@@ -38,12 +44,20 @@ export default class AppbarPage extends Component {
...
               <Appbar title='I am in this drawer!' />
-              <p>{'hello!'}</p>
+              <p>I am a <b>persistent</b> drawer!</p>
+            </Drawer>
+            <Drawer
+              temporary
+              attachment='right'
+              open={this.state.temporaryRight}
+              handleRequestClose={this.toggleTemporaryRight}>
+              <Appbar title='I am a temporary appbar!' />
```
ashwoodall: one more NIT: <em>technically</em> the appbar isn't temporary, but the drawer is

ceciliamvrie: Hmm, there seems to be different app bars depending on the type of drawer that's with it - whether it's permanent or it's temporary 
https://material.io/guidelines/patterns/navigation-drawer.html#navigation-drawer-content

ashwoodall: Is that the subheader component though?

ceciliamvrie: I see what you mean. I'll change it, thanks!

PR #154, opened 2018-02-22
[RFC]style(linting): add stylelint

AriLFrankel:
[Stylelint](https://github.com/stylelint/stylelint) is a styles linting tool with support for and from styled components.

There are handy text editor [plugins](https://github.com/stylelint/stylelint/blob/434cf4660d824e183d94d9897f9d181f64c09172/docs/user-guide/complementary-tools.md#editor-plugins) including VS Code and even _vim_

This PR adds a basic stylelint implementation as described [here](https://www.styled-components.com/docs/tooling)

*Pros*: Linting 👍 
*Cons*: --fix option is experimental, so there's a fair amount of work that would go in to the retroactive linting that would have to happen. Additionally, linting requirements increase the work that goes in to making a contribution to this repo.

My feeling is that this would be a win for this repo for all of the reasons that we lint everywhere else across our code bases. The [lack of a viable --fix option](https://github.com/stylelint/stylelint/issues/3189) would be a pain point until that's resolved.

After the retroactive linting is finished I think it'd be worth adding to the prepush hook

I think that this would be good to add to our other repos as well

### DIFF
--- a/.stylelintrc
+++ b/.stylelintrc
@@ -1,0 +1,8 @@
+{
+  "processors": ["stylelint-processor-styled-components"],
+  "extends": [
+    "stylelint-config-standard",
+    "stylelint-config-styled-components"
+  ],
+  "syntax": "scss"
+}
--- a/package.json
+++ b/package.json
@@ -8,6 +8,7 @@
     "clear": "rm -rf dist/*",
     "build": "yarn clear && babel --out-dir dist src",
     "lint": "eslint src/**",
+    "lint:css": "stylelint {src,pages}/**",
     "dev": "next dev -p 8081",
     "export": "next build && next export -o docs",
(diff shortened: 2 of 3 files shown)


martahj:
This is really neat.

Out of curiosity, what was the motivation behind this? Has inconsistent CSS formatting been a pain point?

AriLFrankel:
I wouldn't say that inconsistent CSS has been a pain point, no. Mostly I saw this on styled components page and thought it looked pretty neat.

While the linting rules are almost all cosmetic at this point, I think that similar to our other code bases there's an opportunity to extend this configuration with more implementation-oriented rules as this repo matures.

martahj:
@AriLFrankel Gotcha. I'm 👍 with giving css linting a try and seeing how it can improve our life 🏖

PR #159, opened 2018-02-23
refactor(Checkbox): remove classNames, export as S-C

petegivens:
- Removed classNames and passed props.className to the outer wrapper to support Component selector pattern

- Refactored icon into a single component. 

- Renamed subcomponents for simplicity and exported those along with the default export to allow for component selector styling.


brad-decker:
@martahj when you've reviewed this could you merge?

PR #160, opened 2018-02-28
feat(Menu): Add base menu component

### LINKED ISSUE #27: Create Menu component
https://material.io/guidelines/components/menus.html

ashwoodall:
First pass at menu component. This component is accessible and
navigable by keyboard events and usable in a form via  hidden
select/option elements

Resolve #27 

Demo: https://docs-lqtopesdba.now.sh

### DIFF
--- a/pages/lists.js
+++ b/pages/lists.js
@@ -6,7 +6,7 @@ import ListItemText from '../src/components/List/ListItemText';
 import MaterialThemeProvider from '../src/theme/ThemeProvider';
 
 const ListsPage = ({ className }) => (
-  <MaterialThemeProvider theme={{ primary: "#03A9F4" }}>
+  <MaterialThemeProvider theme={{ primary: '#03A9F4' }}>
     <div className={className}>
       <h1>List with Avatar</h1>
       <div className="list">
@@ -40,31 +40,34 @@ const ListsPage = ({ className }) => (
       <div className="list">
         <List>
           <ListItem leftAvatar="http://www.wasiofaces.com/wp-content/uploads/San-Diego-Headshots-Photography-041.jpg" avatarSize={60}>
-            <ListItemText primary="Primary Title" secondary={
-              <div>
-                <p>(214) 520-8300</p>
-                <p>5015 Tracy Street Suite 102</p>
-                <p>Dallas, Texas 75205</p>
-              </div>
-            } />
+            <ListItemText
+              primary="Primary Title" secondary={
+                <div>
+                  <p>(214) 520-8300</p>
+                  <p>5015 Tracy Street Suite 102</p>
+                  <p>Dallas, Texas 75205</p>
+                </div>
+              } />
           </ListItem>
           <ListItem leftAvatar="http://www.wasiofaces.com/wp-content/uploads/San-Diego-Headshots-Photography-041.jpg" avatarSize={60}>
-            <ListItemText primary="Primary Title" secondary={
-              <div>
-                <p>(214) 520-8300</p>
-                <p>5015 Tracy Street Suite 102</p>
-                <p>Dallas, Texas 75205</p>
-              </div>
-            } />
+            <ListItemText
+              primary="Primary Title" secondary={
+                <div>
+                  <p>(214) 520-8300</p>
+                  <p>5015 Tracy Street Suite 102</p>
+                  <p>Dallas, Texas 75205</p>
+                </div>
+              } />
           </ListItem>
           <ListItem leftAvatar="http://www.wasiofaces.com/wp-content/uploads/San-Diego-Headshots-Photography-041.jpg" avatarSize={60}>
-            <ListItemText primary="Primary Title" secondary={
-              <div>
-                <p>(214) 520-8300</p>
-                <p>5015 Tracy Street Suite 102</p>
-                <p>Dallas, Texas 75205</p>
-              </div>
-            } />
+            <ListItemText
+              primary="Primary Title" secondary={
+                <div>
+                  <p>(214) 520-8300</p>
+                  <p>5015 Tracy Street Suite 102</p>
+                  <p>Dallas, Texas 75205</p>
+                </div>
+              } />
           </ListItem>
         </List>
       </div>
@@ -74,19 +77,19 @@ const ListsPage = ({ className }) => (
           <ListItem
             withDivider
             leftAvatar="http://www.wasiofaces.com/wp-content/uploads/San-Diego-Headshots-Photography-041.jpg"
-            >
+          >
             <ListItemText primary="Primary Title" secondary="This is the secondary line" />
           </ListItem>
           <ListItem
             withDivider
             leftAvatar="http://www.wasiofaces.com/wp-content/uploads/San-Diego-Headshots-Photography-041.jpg"
-            >
+          >
             <ListItemText primary="Primary Title" secondary="This is the secondary line" />
           </ListItem>
           <ListItem
             withDivider
             leftAvatar="http://www.wasiofaces.com/wp-content/uploads/San-Diego-Headshots-Photography-041.jpg"
-            >
+          >
             <ListItemText primary="Primary Title" secondary="This is the secondary line" />
           </ListItem>
         </List>
--- a/pages/menus.js
+++ b/pages/menus.js
@@ -1,0 +1,68 @@
+import React, { Component } from 'react';
+import Button from '../src/components/Button';
+import MenuItem from '../src/components/Menu/MenuItem';
+import Menu from '../src/components/Menu/Menu';
+import MaterialThemeProvider from '../src/theme/ThemeProvider';
+
+const StandAloneMenu = Menu.extend`
+  position: relative;
+`;
+
+class MenusPage extends Component {
+  state = {
+    open: false,
+    value: '',
+  };
+  
+  handleClick = () => {
+    this.setState({ open: !this.state.open });
+  };
+
+  handleSelect = (selectedItem) => {
+    const item = selectedItem;
+    this.setState({ open: false, value: item || this.state.value });
+  };
+
+  handleClose = () => {
+    this.setState({ open: false });
+  };
+
+  render() {
+    const menuItems = [
+      { text: 'Un', onClick: () => this.handleSelect() },
+      { text: 'Deux', onClick: () => this.handleSelect() },
+      { text: 'Trois', onClick: () => this.handleSelect() },
+      { text: 'Quatre', onClick: () => this.handleSelect() },
+    ];
+
+    return (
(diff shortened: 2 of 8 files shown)


brad-decker:
The UI looks great. Will do a code review soon 👍

AriLFrankel:
Haven't looked at the code yet, but I do get this error when I try to use enter to select an item
screen shot 2018-03-01 at 9 16 29 am

pages/menus.js
```diff
@@ -0,0 +1,83 @@
...
+  
+  handleClick = () => {
+    this.setState({ open: !this.state.open });
+  };
+
+  handleSelect = (selectedItem) => {
+    const item = selectedItem;
+    this.setState({ open: false, value: item });
+  };
+
```
danielstclair: We've talked about this offline but I wanted to open source this discussion to the other devs. Is there a use case for a menu that doesn't open and close or we don't use MenuSelect and MenuOption? I understand wanting to give the developer a lot of control over this but this seems like a use case so common that it could be baked in to the component. I'd love to hear other people's thoughts on this.

ashwoodall:
@AriLFrankel that's really weird; I'm not getting that error when I load the page. Let me investigate 🕵️‍♀️

ashwoodall:
Hey @danielstclair, currently you do have to focus with tab on the component, but then it is navigable with the up/down arrows.

src/components/Menu/HiddenSelect.js
```diff
@@ -0,0 +1,23 @@
+import React from 'react';
+import styled from 'styled-components';
+
+const HiddenSelectEl = styled.select`
+  visibility: hidden;
```
AriLFrankel: [q] What's the reason for having both `visibility: hidden` and `display:none` here?

src/components/Menu/HiddenOption.js
```diff
@@ -0,0 +1,13 @@
+import React from 'react';
+import styled from 'styled-components';
+
+const HiddenOptionEl = styled.option`
+  visibility: hidden;
+  display: none;
```
AriLFrankel: [q] same question as below

pages/menus.js
AriLFrankel: Can you help me understand how you envision the API looking differently with regard to open/close functionality, and without MenuSelect and MenuOption?

pages/lists.js
```diff
@@ -6,7 +6,7 @@ import ListItemText from '../src/components/List/ListItemText';
 import MaterialThemeProvider from '../src/theme/ThemeProvider';
 
 const ListsPage = ({ className }) => (
-  <MaterialThemeProvider theme={{ primary: "#03A9F4" }}>
+  <MaterialThemeProvider theme={{ primary: '#03A9F4' }}>
```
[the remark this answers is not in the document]
ashwoodall: yep. I can revert if this is annoying

src/components/Menu/HiddenSelect.js
danielstclair: @AriLFrankel This is for better a11y with screen readers. Unfortunately, limitations in html 5 don't allow us to completely manipulate select and option tags to get this material interaction up to spec.

ashwoodall: This is a hidden select element for accessibility (screen readers) and for use in submitting forms. Apparently it's impossible to add custom styling to select/option elements, which is why the styling is handled by visible ul and li elements here (hidden by aria for screenreaders)

pages/lists.js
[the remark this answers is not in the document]
brad-decker: no, the single quotes in an object literal are correct

src/components/Menu/HiddenSelect.js
AriLFrankel: I totally get the need to have them hidden, but can you help me understand the need for the visibility property over the top of the `display: none`? 

I might be totally missing the point here, but my understanding is that the only difference between `display: none` and `visibility:hidden` is that the later still allows the element to take up space, so that including `display: none` renders `visibility:hidden` redundant. https://codepen.io/anon/pen/VQRYyg

danielstclair: ahh gotcha @AriLFrankel sorry I misread the question.

ashwoodall: Ah sorry, me too. Updating this now.

pages/menus.js
danielstclair: @AriLFrankel Sorry I didn't see this for some reason.
The idea would be that we pass in the options data as a prop and perhaps a few other things to accommodate for other use cases. Maybe something like this:

```
<Menu
  options={arrayOfOptions}
  initialValue="SomeValue"
  cb={optionalFunc}
  />
```

AriLFrankel: @danielstclair thanks, no problem! It looks like I accidentally deleted my comment 5 minutes ago _shrug_ 

Regarding options, I see what you mean now. That's interesting. I'd be 👍 on that API

Regarding open/close, what are your thoughts on how that might be different?

danielstclair: @AriLFrankel Not sure about different in terms of functionality as much as where it is. If a menu item is always something that should be opened or closed my thought is to write the functionality into the component itself instead of the page/parent component that's using the menu option as that would reduce the amount of rewrite for an interaction that in theory should always happen.

Downside to these propositions are if there are some sort of custom interactions/use cases the the developer is trying to account for. Having too much functionality baked in would make it difficult to write out these custom use cases. Of course, this entire argument is null if the open/close functionality is not something that should always happen.

Again, I'm really just wanting some thoughts here. I'm totally fine with leaving it the way it is which would give all the power to the end user with the expense of writing out some interactions that may or may not be deemed as necessary to the component.

AriLFrankel: @danielstclair for me it makes sense to control the open/close from without the menu, just because this is something that I imagine the developer will want to control, as you mentioned. While I do think that that interaction will probably almost always happen, I think that the specifics of that interaction would be highly dependent on the context in which it's written. Per the spec, "A menu is a temporary piece of material that appears upon interaction with a button, action, pointer, or other control" Given the many options, I think we should be as flexible as possible.

ashwoodall:
Okay @AriLFrankel @danielstclair @brad-decker,

I've significantly updated this component. This menu now supports:

- accepting either MenuList/MenuListItems (like the current List/ListItem pattern) as children OR accepting an array of objects as menuItems props like 
```    
const menuItems = [
      { text: 'Un', onClick: () => this.handleSelect() },
      { text: 'Deux', onClick: () => this.handleSelect() },
      { text: 'Trois', onClick: () => this.handleSelect() },
      { text: 'Quatre', onClick: () => this.handleSelect() },
    ]; 
```
- improved up/down array interactions (automatically focuses on the first item when opened so that the arrow keys are accessible without having to tab to focus into the menu
- esc to close
- click outside of the menu to close 
- stripped out the select/option elements and will implement that hidden element logic later for implementing a separate select/dropdown menu
- Accessibility for screenreaders is now handled by the ARIA role: menu attribute


Check out the new changes here - https://docs-xzvjtoefgw.now.sh

pages/menus.js
```diff
@@ -0,0 +1,68 @@
...
+    ];
+
+    return (
+      <MaterialThemeProvider theme={{ primary: '#03A9F4' }}>
+        <div style={{ margin: '60px 16px' }}>
+          <h1>Menus</h1>
+          <h2>Standalone Menu</h2>
+          <p>Menus can accept MenuSelect and MenuOption SMC components as children,
+             or can accept an array of objects as seen here - </p>
+          <StandAloneMenu open onClose={() => {}} menuItems={menuItems} />
```
AriLFrankel: [q] Looks like you have to pass in an onClose handler. What's the reason for making it not optional?

src/components/Menu/Menu.js
```diff
@@ -0,0 +1,60 @@
...
+  }
+  
+  componentWillUnmount() {
+    document.removeEventListener('mousedown', this.handleClick, false);
+  }
+
+  handleClick = (event) => {
+    if (this.menu.contains(event.target)) return;
+
+    this.props.onClose();
```
AriLFrankel: [q] same as above

AriLFrankel:
@ashwoodall Awesome job on adding the support for passing in an array of options 👍 I would have totally been comfortable approving this as is - nice addition!

Left just one q

pages/menus.js
ashwoodall: These props now 100% more optional!

src/components/Menu/Menu.js
```diff
@@ -0,0 +1,59 @@
...
+    document.addEventListener('mousedown', this.handleClick, false);
+  }
+  
+  componentWillUnmount() {
+    document.removeEventListener('mousedown', this.handleClick, false);
+  }
+
+  handleClick = (event) => {
+    if (this.menu.contains(event.target)) return;
+    this.props.onClose();
```
danielstclair: look like we have one left that needs to be checked for

ashwoodall: The component has a default parameter defined if it's not given an onClose prop

`const { className, children, menuItems, open, onClose = ({}) } = this.props;`

src/components/Menu/Menu.js
```diff
@@ -0,0 +1,59 @@
...
+  }
+
+  render() {
+    const { className, children, menuItems, open, onClose = ({}) } = this.props;
```
AriLFrankel: [q] I might be missing something, but it seems like that would still throw an error, wouldn't it? `({})()` throws

ashwoodall: I'm not getting any errors? Should just return an empty object

src/components/Menu/Menu.js
ashwoodall: well that handleClick is to handle closing on clicks outside of the menu, so it's open and not intended to close it shouldn't matter

src/components/Menu/Menu.js
```diff
@@ -0,0 +1,59 @@
...
+    const renderMenuItems = open && menuItems && menuItems.length > 0;
+    const renderChildren = open && !menuItems;
+
+    return (
+      <div className={`${className} smc-Menu`} ref={(ref) => {this.menu = ref;}}>
+        {renderChildren &&
+          <MenuList onKeyDown={this.handleKeyDown} onClose={onClose}>
+            {children}
+          </MenuList>
+        }
```
petegivens: [Q] Is the intention to allow a combination of a list of menuItems plus any children added? Seems like if the user passed in a `menuItems` prop and also provided children, it would render two separate MenuLists.

ashwoodall: Yeah, I think eventually we will need to adopt one pattern

PR #161, opened 2018-03-01
Features: Appbar Navigation, FlexGrid

brad-decker:
### Links
https://docs-oiaijpnyfw.now.sh/appbar
https://docs-oiaijpnyfw.now.sh/flex-grid

### Flex Grid
[Material Flex Grid Specification](https://material.io/guidelines/layout/responsive-ui.html)

Most everything in material design adheres to this flex-grid specification including gutters and margins, etc. The flexible grid specification is a layout of columns and gutters that help fit components together in the style deemed appropriate. My hope is that eventually everything fits into columns and rows in userland and sets the gutters this way so that components just play nicely together regardless of composition.

What i've done so far:
1. Added respondable (our own package) for detecting screen size changes.
2. Added some theme options for specifying screen sizes and priority of sizes for respondable.
3. Added create-react-context (context api polyfill by james kyle)
4. Created a 'ScreenSize' provider with this context that initiates respondable using the theme options.
5. ScreenSize Provider is attached in the ThemeProvider component so all installations of our library will initialize a respondable instance and provide the context of the current screensize to our components.
6. Created Row and Column components that allow for responsive sizing, ordering, offsets as well as common use cases for rows such as horizontal and vertical positioning. 
7. Created a Row Context Provider that can be used to get the current Row's props inside of columns that are in a row. This allows columns to adjust their styling based on the configuration of the row they are within. 
8. Added the DESKTOP breakpoints from the material spec. 
9. Created a demo page for some of the configurations.

There is more to do to make this really robust but this is a good start. I'd like to be able to adjust the gutters and what not based on screensize as well. 

### App Bar Navigation
This is not actually material spec but a nice to have feature that we will use at Merlin Labs. This new feature allows passing a 'navigation' child into the Appbar that will render a navigation menu. It also adds support for the 'dense' and 'double' props on Appbar per material spec. i've got more work to do here but i don't wanna block others.

**Breaking Change?** - Appbar has changed significantly and I know we are using it already on the cms side @danielstclair @ashwoodall @BennettStaley please be thorough and let me know what will make it not such a PITA to implement the update.


next.config.js
```diff
@@ -2,6 +2,23 @@ module.exports = {
...
+      '/divider': { page: '/divider' },
+      '/drawer': { page: '/drawer' },
+      '/gridlists': { page: '/gridlists' },
+      '/lists': { page: '/lists' },
+      '/sliders': { page: '/sliders' },
+      '/snackbar': { page: '/snackbar' },
+      '/switches': { page: '/switches' },
+      '/tables': { page: '/tables' },
+      '/tabs': { page: '/tabs' },
+      '/text-fields': { page: '/text-fields' },
```
martahj: The now app routing issue! You've fixed it! 🌟

src/components/Appbar/Navigation.js
```diff
@@ -0,0 +1,79 @@
...
+      width: element.clientWidth,
+      offset: element.offsetLeft,
+    }));
+  }, 100);
+
+  render() {
+    return (
+      <div data-smc="AppbarNavigation" className={this.props.className}>
+        {this.props.children}
+        <InkBar width={this.state.width} offset={this.state.offset} />
```
martahj: [Q] Seems like this would be a good use case for the tabs - is there a reason they didn't work in here?

src/components/Appbar/Navigation.js
```diff
@@ -0,0 +1,79 @@
...
+class NavigationContainer extends Component {
+  state = {
+    width: 0,
+    offset: 0,
+    active: this.props.active,
+  };
+
+  componentDidMount() {
+    this.resizeInkBar();
+    window.addEventListener('resize', this.resizeInkBar);
```
martahj: [Q] Thoughts on removing the event listener on `componentWillUnmount`?

brad-decker: Definitely should do that yes.

src/components/Appbar/Navigation.js
brad-decker: Tabs have highlight effects on active tabs, they also should always be equal width, and a number of other things that don't make them a good fit for this. There's also a semantic meaning to a tab versus a navigation item.

src/components/Appbar/Navigation.js
brad-decker: @martahj resolved

martahj:
@brad-decker Did you mean to share this link https://docs-qutpczuhxp.now.sh/flex-grid for the flex-grid example, or is the one you posted more up to date? I'm getting a "cannot read property 'horizontal' of undefined" error at this url https://docs-gklyasqvpx.now.sh/flex-grid

brad-decker:
@martahj thanks for pointing that out. Yeah i redid something after i posted the url. Fixing now

brad-decker:
@Nathan-Schwartz mentioned it'd be helpful to have more comments around the respondable / context api stuff. when you review this if you could drop a line on anything that you'd like explained more with inline comments that'll help me figure out the best places to comment

BennettStaley:
screen shot 2018-03-05 at 1 39 16 pm
It appears that the appbar doesn't go over the ink-bars. They also seem to not be tied to their container and jello-wiggle back to place after resizing the page.

brad-decker:
@BennettStaley the inkbar is actually positioned relative to to the navigation menu and is positioned via the offset of the current active menu item. Because that menu item changes offset position when screen resizing occurs, the inkbar updates its location after the resize event. I think that could probably be improved, but for now i'm inclined to land it and work on improving it in another PR. Thoughts?

brad-decker:
Will definitely fix the z-index issue though

src/components/Appbar/Appbar.js
```diff
@@ -0,0 +1,66 @@
...
+  margin: 0 12px;
+  height: ${props => props.height}px;
+  display: flex;
+  align-items: center;
+  flex: 1;
+  ${typography('title')}
+`;
+
+const AppbarComponent = ({ className, title, navIcon, children, dense, double = false }) => (
+  <div className={className} data-smc="Appbar">
```
danielstclair: I'm wondering about the appbar component. Since it's something that would always be at the top of it's section and also contain either some kind of heading or navigational content, it might might sense to make it a header instead of div.

src/components/Appbar/Navigation.js
```diff
@@ -0,0 +1,74 @@
...
+    if (!element) return;
+    this.setState(() => ({
+      width: element.clientWidth,
+      offset: element.offsetLeft,
+    }));
+  }, 100);
+
+  render() {
+    return (
+      <React.Fragment>
```
danielstclair: I think this should probably be a nav element instead of a fragment.

src/components/Appbar/Navigation.js
```diff
@@ -0,0 +1,74 @@
...
+        {this.props.children}
+        <InkBar width={this.state.width} offset={this.state.offset} />
+      </React.Fragment>
+    );
+  }
+}
+
+export const InkBar = styled.div`
```
danielstclair: A div totally makes sense here. It might not be an issue but maybe we should add
```aria-hidden=true``` to the jsx to make sure screen readers don't catch it.

File: src/components/FlexGrid/Column.js [hunk 7]
```diff
@@ -0,0 +1,147 @@
...
+  ${flexSizeMixin}
+  ${orderMixin}
+`;
+
+const ColumnComponent = ({ className, children, ...props }) => (
+  <ScreenSizeConsumer>
+    {({ screenSize, lowerScreenSize }) => (
+      <RowConsumer>
+        {row => (
+          <SizedColumn
```
danielstclair: So I think this might be tricky. A div is always going to play nicely with these styles but a column could be a few different things than a div (li, section, article, etc). I'm wondering if this should have another prop that we can pass in from the parent component that would give us some more control over the aria-role? What do you think?

src/components/FlexGrid/Row.js
```diff
@@ -0,0 +1,109 @@
...
+    if (this.props.stretch !== nextProps.stretch) {
+      newState.stretch = nextProps.stretch;
+    }
+    if (Object.keys(newState).length) this.setState(newState);
+  }
+
+  render() {
+    return (
+      <Provider value={this.state}>
+        <div data-smc="Row" className={this.props.className}>
```
danielstclair: Same comment as above. I think it might be nice to have greater control over the ARIA role of this element.

src/contexts/ScreenSizeContext.js
```diff
@@ -0,0 +1,59 @@
...
+          ...this.state,
+        }}
+      >
+        {this.props.children}
+      </Context.Provider>
+    );
+  }
+}
+
+export const ScreenSizeContext = withTheme(ScreenSizeContextBase);
```
danielstclair: This whole thing is just really cool.

src/components/FlexGrid/Column.js
brad-decker: My opinion is that the column should only be used as a container around the above things, but I am not versed in this facet of accessibility. The intended use of the column component is to be a container:
```jsx
<Column xs={{ size: 12 }} sm={{ size: 6 }}>
  <article>...</article>
</Column>
```

src/components/Appbar/Navigation.js
brad-decker: cool!

src/components/Appbar/Navigation.js
brad-decker: Fragment is a react component that does not render a html node. its for returning a set of root level components without a parent.

src/components/Appbar/Appbar.js
brad-decker: I'm fine with header.

src/components/Appbar/Navigation.js
danielstclair: Understood. The reason I mentioned it is so that it would always act as a nav element. The alternative would be simply wrapping this in a nav element when consumed which would work.
```<nav><Navigation /></nav>``` vs ```<Navigation />```

Comment (brad-decker) [reply src/components/Appbar/Navigation.js]: The problem that I solved by using Fragment here is that the layout of the AppBar is being handled inside the AppBar itself. Whether or not you use the Navigation you can still pass children into AppBar (think like buttons and what not that you want positioned to the right on the AppBar).

I Get what you're saying but its gonna require more research on how to pull it off.

File: src/components/FlexGrid/Column.js (hunk 7 above)
Comment (danielstclair) [reply src/components/FlexGrid/Column.js]: I think the edge case I'm concerned with is if a column is wrapping itself around a ul which is wrapped around a row. The rendered html would be
```
<div data-smc="row">
  <ul>
    <div>
      <li />
    </div>
  </ul>
</div>```
In that example a screen reader would not be able to associate the li to the ul as their direct children should always be an li.

brad-decker: I think we should elevate this into a new issue. I don't disagree with you, but i'm pretty sure this edge case would be better suited with implementation in user land. Flex Grid is meant for layout styling not creating a GridList (we have a separate component for that that probably needs some ARIA TLC).

src/components/Appbar/Navigation.js
brad-decker: Added

src/components/Appbar/Appbar.js
brad-decker: Fixed!

brad-decker:
@BennettStaley resolved your feedback item regarding z-index issue. @martahj ready for re-review

BennettStaley:
The z index looks fixed. Have you thought about when you might be implementing the fix for the jelly ink-bar?

brad-decker:
Discussed the inkbar item with @BennettStaley in slack. I think it needs improving but outside of this PR

BennettStaley:
Pinging @martahj before we merge, since she had questions.

PR #162, opened 2018-03-02
Pagination table

martahj:
* Creates a table with pagination
* Adds material icons

PR app can be found [here](https://docs-sutmmvifij.now.sh)

### DIFF
--- a/pages/icons.js
+++ b/pages/icons.js
@@ -1,0 +1,104 @@
+import React, { Fragment } from 'react';
+import styled from 'styled-components';
+import MaterialThemeProvider from '../src/theme/ThemeProvider';
+import {
+  Icon,
+  ArrowDropDown,
+  KeyboardArrowLeftIcon,
+} from '../src/icons/icons';
+
+const availableIcons = [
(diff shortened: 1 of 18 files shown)


pages/icons.js
```diff
@@ -0,0 +1,104 @@
+import React, { Fragment } from 'react';
```
ceciliamvrie: [Q] Is your intention to use a Fragment in the future?

pages/tables.js
```diff
@@ -38,37 +39,143 @@ const data = [
...
+              <li>
+                key: string // React key, must be unique
+              </li>
+              <li>
+                [columnKey: string]: any // property should match the field key
+              </li>
+            </ul>
+          </li>
+          <li>
+            rowsPerPage?: number
```
ceciliamvrie: [Q] Is this an idiomatic ternary operator in a list element?

martahj: It's using flow's notation to indicate that the rowsPerPage prop is optional, but if it is passed it should be a number

PR #166, opened 2018-03-02
feat(Tabs): Allow controlled

martahj:
Implements controlled tabs

Now link https://docs-totoujivtt.now.sh

### DIFF
--- a/pages/tabs.js
+++ b/pages/tabs.js
@@ -2,12 +2,49 @@ import React from 'react';
 import styled from 'styled-components';
 import { Tabs, Tab, TabsContainer, TabNavigationContainer } from '../src/index';
 import MaterialThemeProvider from '../src/theme/ThemeProvider';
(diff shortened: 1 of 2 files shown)


pages/tabs.js
```diff
@@ -2,12 +2,49 @@ import React from 'react';
...
+  handleButtonClick = () => {
+    this.setState(({ selectedIndex }) => {
+      if (selectedIndex > 1) return { selectedIndex: 0 };
+      return { selectedIndex: selectedIndex + 1 };
+    })
+  }
+
+  render() {
+    return (
+      <React.Fragment>
```
ceciliamvrie: [Q] What's your reason for using a fragment here?

martahj: So we don't have to wrap all this in a div or generate arbitrary unique keys for an array return

PR #167, opened 2018-03-02
Implement dropdown menu

ceciliamvrie:
#Implemented a new textfield to be used in the property details page. 
Added support for multiple browsers, as well as a11y. 

link => https://docs-wnyapgsmik.now.sh
spec link => https://material.io/guidelines/components/text-fields.html#text-fields-layout

### DIFF
--- a/pages/text-fields.js
+++ b/pages/text-fields.js
@@ -38,7 +38,7 @@ class TextFieldPage extends PureComponent {
           <TextFieldWithBottomMargin defaultValue={'Default Value'} />
           <TextFieldWithBottomMargin
             floatingLabelText={'floating label'}
-          />
+          /> 
           <TextFieldWithBottomMargin
             hintText={'with hint text'}
             floatingLabelText={'floating label'}
@@ -114,6 +114,28 @@ class TextFieldPage extends PureComponent {
             onReset={() => this.setState({ shouldReset: false })}
           />
           <Button onClick={() => this.setState({ shouldReset: true })}>Reset</Button>
+          <h2>Dropdown Menu</h2>
+          <TextFieldWithBottomMargin
+              floatingLabelText="Dropdown Menu"
+              options={["option1", "option2"]}
+          />
+          <h4>No Selected Default Option</h4>
+          <TextFieldWithBottomMargin
+              floatingLabelText="Dropdown Menu"
+              options={["option1", "option2"]}
+              defaultOption={"select an option"}
+          />
+          <h4>Selected Default Option</h4>
+          <TextFieldWithBottomMargin
+              floatingLabelText="Dropdown Menu"
+              options={["option0", "option1", "option2"]}
+              defaultOption={"option2"}
+          />
+          <h4>Invalid Options Passed In</h4>
+          <TextFieldWithBottomMargin
+              floatingLabelText="Dropdown Menu"
+              options={"invalid option"}
+          />
         </div>
       </MaterialThemeProvider>
     );
--- a/src/components/DropdownMenu.js
+++ b/src/components/DropdownMenu.js
@@ -1,0 +1,119 @@
+import styled from 'styled-components';
+import React, { Component, Fragment } from 'react';
+import MenuItem from './Menu/MenuItem';
+import Menu from './Menu/Menu';
+import { ArrowDropDown } from './../icons/icons';
+
+const Dropdown = styled.select`
+  border-color: transparent;
+  background-color: transparent;
+  color: black;
+  width: 100%;
+  position: absolute;
+  z-index: 1;
+  font-size: 17px;
+  -webkit-appearance: button; /* hide default arrow in chrome OSX */
+
+  &:focus {
+    outline: none;
+  }
+`;
+
+const HiddenOption = styled.option`
+  display: none;
+`;
+
+const Symbol = styled(ArrowDropDown)`
+  width: 20px;
+  height: 20px;
+  position: absolute;
+  right: 0;
+  bottom: 5px;
+  fill: #726969;
+`;
+
+
+export default class DropdownMenu extends Component {
+  state = {
+    isOpen: false,
+    options: this.props.options || [],
+    selected: this.props.defaultOption || 'select one',
+    isChrome: undefined,
+  }
+  
+  /* eslint-disable react/no-did-mount-set-state */
+  /*
+   * Because Next.js executes its code server-side first,
+   * using navigator and window will result to a reference error
+   * outside of the componentDidMount method. Therefore, setting
+   * the state inside of here is the only way to keep track of what
+   * browser is being used on the client side.
+   *
+   * https://github.com/zeit/next.js/wiki/FAQ
+   */
+
+  componentDidMount() {
+    if (navigator) {
+      const isChrome = Boolean(/Chrome/.test(navigator.userAgent) && /Google Inc/.test(navigator.vendor));
+      this.setState({ isChrome });
+    }
+  }
+  /* eslint-enable react/no-did-mount-set-state */
+   
+  onSelectMenuItem = (option): void => {
+    this.setState({ selected: option });
+  }
+
+  toggleSelect = (e): boolean => {
+    e.preventDefault();
+    this.setState({ isOpen: !this.state.isOpen });
+  }
+
+  render() {
+    const { options, isOpen, selected, isChrome } = this.state;
+    return (
+      <div onClick={this.toggleSelect}>
+        {isChrome === false && (
+          <Dropdown
+            defaultValue={selected}
+          >
+            {options.map(option => (
+              <option key={option}>
+                {option}
+              </option>
+            ))}
+          </Dropdown>
+        )}
+        {isChrome === true && (
+          <Fragment>
+            <Dropdown
+              value={selected}
+              hidden={isOpen}
+            >
+              {options.map(option => (
+                <HiddenOption key={option}>
+                  {option}
+                </HiddenOption>
+              ))}
+            </Dropdown>
+            <Menu
+              aria-hidden="true"
+              value={selected}
+              open={isOpen}
+            >
+              {options.map(option => (
+                <MenuItem
+                  key={option}
+                  onClick={() => this.onSelectMenuItem(option)}
+                >
+                  {option}
+                </MenuItem>
+              ))}
+            </Menu>
+          </Fragment>
+        )}
+        <Symbol />
+      </div>
+    );
+  }
+}
--- a/src/components/TextField.js
+++ b/src/components/TextField.js
@@ -1,6 +1,7 @@
 /* eslint-disable no-confusing-arrow */
 import styled, { css } from 'styled-components';
 import React, { PureComponent } from 'react';
+import DropdownMenu from './DropdownMenu';
 
 class TextFieldComponent extends PureComponent {
   state = {
@@ -45,6 +46,11 @@ class TextFieldComponent extends PureComponent {
   };
 
   render() {
+    const hasValidOptions = this.props.options && (
+      Array.isArray(this.props.options) &&
+      this.props.options.length > 0
+    );
+
     const hasError = Boolean(this.state.error || this.props.error || this.props.errorText);
     return (
       <div className={`${this.props.className} smc-text-field-container`}>
@@ -58,7 +64,12 @@ class TextFieldComponent extends PureComponent {
           floatingLabelStyle={
             hasError ? this.props.floatingLabelErrorStyle : this.props.floatingLabelStyle
           }
-          floating={this.state.focus || this.props.hintText || this.state.text.length}
+          floating={this.state.focus ||
+            this.props.hintText ||
+            this.props.options ||
+            this.props.defaultOption ||
+            this.state.text.length
+          }
         >
           {this.props.floatingLabelText || ''}
           {this.props.required ? '*' : ''}
@@ -67,15 +78,23 @@ class TextFieldComponent extends PureComponent {
             show={this.props.required}
             requiredStarStyle={this.props.requiredStarStyle} /> */}
         </FloatingLabel>
-        <HintText
-          className={'smc-text-field-hint-text'}
-          hintTextStyle={this.props.hintTextStyle}
-          hasPrefix={this.props.prefix}
-          error={hasError}
-          show={!this.props.defaultValue && !this.state.text.length && !this.props.value}
-        >
-          {this.props.hintText}
-        </HintText>
+        {hasValidOptions && (
+          <DropdownMenu
+            defaultOption={this.props.defaultOption}
+            options={this.props.options}
+          />
+        )}
+        {!this.props.options && (
+          <HintText
+            className={'smc-text-field-hint-text'}
+            hintTextStyle={this.props.hintTextStyle}
+            hasPrefix={this.props.prefix}
+            error={hasError}
+            show={!this.props.defaultValue && !this.state.text.length && !this.props.value}
+          >
+            {this.props.hintText}
+          </HintText>
+        )}
         {this.props.helperText && (
           <HelperText
             className={'smc-text-field-helper-text'}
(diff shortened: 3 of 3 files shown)


ceciliamvrie:
I believe so!

ceciliamvrie:
@danielstclair Yeah, I believe that can happen! Hey @ashwoodall, do you think I can use the menu component for my dropdown menu?

ashwoodall:
@ceciliamvrie Absolutely! I think that would definitely bring it closer to material design. I think what you have here is a good first step solution if you guys need this dropdown immediately. Eventually the Menu ul/li elements would need to map to a hidden select/option for accessibility though

src/components/DropdownMenu.js
```diff
@@ -0,0 +1,92 @@
...
+  fill: #726969;
+`;
+
+export default class DropdownMenu extends Component {
+  constructor(props) {
+    super(props);
+    this.state = { isOpen: false, selected: '', options: [] };
+  }
+   
+  componentDidMount = (): void => {
```
martahj: [Q] What's the reason for having this code in `componentDidMount`?

src/components/DropdownMenu.js
```diff
@@ -0,0 +1,92 @@
...
+    const { options } = this.props;
+    if (options) {
+      this.setState({ selected: options[0], options });
```
ashwoodall: Calling setState in componentDidMount will trigger a re-render. Is that avoidable here?

https://reactjs.org/docs/react-component.html#componentdidmount

src/components/DropdownMenu.js
```diff
@@ -0,0 +1,92 @@
...
+    } else {
+      const optionsArray = ['Default Option'];
+      this.setState({ selected: optionsArray[0], options: optionsArray });
+    }
+  }
+
```
ashwoodall: I actually think you can get away with getting rid of the constructor and componentDidMount if you write state as 

```
  state = {
    isOpen: false,
    selected: this.props.options[0] || 'Default Option',
    options: this.props.options || ['Default Option'],
  }
```

src/components/DropdownMenu.js
```diff
@@ -0,0 +1,92 @@
...
+    this.setState({ isOpen: !this.state.isOpen });
+    return false;
+  }
+
+  render() {
+    const { options, isOpen, selected } = this.state;
+    return (
+      <div onClick={this.toggleSelect}>
+        <Dropdown
+          value={selected}
```
ashwoodall: Does this need to be `defaultValue={selected}`?

src/components/DropdownMenu.js
ceciliamvrie: I want there to be an option selected by default for every dropdown menu. I have this code here to check if there are any options already coming in to the dropdown. If there are not any, then I set a default array of options to be passed in. Do you think it's possible for me to do this inside the constructor?

src/components/DropdownMenu.js
ceciliamvrie: 👍

src/components/DropdownMenu.js
ceciliamvrie: Yeah! Thanks for that, I was trying to do that inside of the constructor but it wasn't working as I expected. Do you know why it's good to leave out the constructor? @ashwoodall

src/components/DropdownMenu.js
ceciliamvrie: Hmm, I don't think I need it since I'm already setting a default value to the state.

src/components/DropdownMenu.js
```diff
@@ -0,0 +1,92 @@
...
+          hidden={isOpen}
+        >
+          {options.map(option => (
+            <HiddenOption key={option}>
+              {option}
+            </HiddenOption>
+          ))}
+        </Dropdown>
+        <Menu
```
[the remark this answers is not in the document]
ceciliamvrie: Cool! Is that just an attribute?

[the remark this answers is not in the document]
ashwoodall: I believe so!

ceciliamvrie:
@ashwoodall @martahj Ready to review!

src/components/DropdownMenu.js
ashwoodall: It allows you to use declarative ES7 syntax for property initializers. 

https://stackoverflow.com/questions/35662932/react-constructor-es6-vs-es7

pages/text-fields.js
```diff
@@ -114,6 +114,12 @@ class TextFieldPage extends PureComponent {
             onReset={() => this.setState({ shouldReset: false })}
           />
           <Button onClick={() => this.setState({ shouldReset: true })}>Reset</Button>
+          <h2>Dropdown Menu</h2>
+          <TextFieldWithBottomMargin
+              floatingLabelText="Dropdown Menu"
+              options={["option1", "option2", "option3"]}
+              dropdown
```
martahj: Not a blocker, but given that the dropdown is the only textfield type that would use an array of options, what do you think about automatically making it a dropdown if the user passes an options prop? This would also mean you wouldn't have to handle the case of a dropdown with options = undefined

src/components/DropdownMenu.js
```diff
@@ -0,0 +1,84 @@
...
+import Menu from './Menu/Menu';
+import { ArrowDropDown } from './../icons/icons';
+
+const Dropdown = styled.select`
+  border-color: transparent;
+  background-color: transparent;
+  color: black;
+  width: 100%;
+  position: absolute;
+  z-index: 1000;
```
martahj: [Q] Does it need to be this high?

src/components/DropdownMenu.js
```diff
@@ -0,0 +1,84 @@
...
+  right: 0;
+  bottom: 5px;
+  fill: #726969;
+`;
+
+export default class DropdownMenu extends Component {
+  state = {
+    isOpen: false,
+    selected: this.props.options[0] || 'Default Option',
+    options: this.props.options || ['Default Option'],
```
martahj: I think lines 38-39 will lead to some unexpected behavior if the user passes anything except an array of length > 0

*Case 1:* User passes an array of length > 0
```js
state.selected = options[0]
state.selected = options
```
No issues here

*Case 2*: Users passes no options
I think this will create a 'Cannot ready property 0 of undefined' error or something similar

*Case 3*: User passes an array of length 0
```js
state.selected = 'Default Option'
state.options = []
```
`state.selected` is set to a value that does not exist in `state.options`

src/components/TextField.js
```diff
@@ -67,15 +72,20 @@ class TextFieldComponent extends PureComponent {
...
-          className={'smc-text-field-hint-text'}
-          hintTextStyle={this.props.hintTextStyle}
-          hasPrefix={this.props.prefix}
-          error={hasError}
-          show={!this.props.defaultValue && !this.state.text.length && !this.props.value}
-        >
-          {this.props.hintText}
-        </HintText>
+        {this.props.dropdown && (
+          <DropdownMenu options={this.props.options} />
```
martahj: A potential future enhancement might be to allow users to pass in the value and selection handler callbacks so the menu can be controlled by the parent

pages/text-fields.js
ceciliamvrie: I like that idea, but I think it's a little implicit. Do you think that it's going to be intuitive for other people? @martahj

src/components/DropdownMenu.js
ceciliamvrie: Yeah, this totally doesn't need to be that high.

src/components/DropdownMenu.js
```diff
@@ -0,0 +1,100 @@
...
+  position: absolute;
+  right: 0;
+  bottom: 5px;
+  fill: #726969;
+`;
+
+export default class DropdownMenu extends Component {
+  state = {
+    isOpen: false,
+    selected: this.props.options[0],
```
danielstclair: I feel like the state.selected needs just a little more work here. Right now we're assuming that the "defaultSelected" should always be the first options. I think that should be something that's explicitly controlled. Also we should have the ability to control whether something isn't selected initially.

Default selected example: I have a list of states as options and I want the default selected state to be Texas.

No default selected example: I have a list of states and I don't want one to be selected by default and want it to say "Please select a state" instead.

I think both of these edge cases would be handled well with a defaultSelected prop. This would also take care of avoiding errors when someone doesn't pass an options array.

src/components/DropdownMenu.js
```diff
@@ -0,0 +1,100 @@
...
+
+  toggleSelect = (e): boolean => {
+    e.preventDefault();
+    this.setState({ isOpen: !this.state.isOpen });
+    return false;
+  }
+
+  render() {
+    const { options, isOpen, selected } = this.state;
+    const isChrome = /Chrome/.test(navigator.userAgent) && /Google Inc/.test(navigator.vendor);
```
martahj: Very cool

src/components/TextField.js
```diff
@@ -45,6 +46,11 @@ class TextFieldComponent extends PureComponent {
   };
 
   render() {
+    const optionsCorrect = this.props.options && (
+      Array.isArray(this.props.options) ||
+      this.props.options.length < 1
```
martahj: Not sure I understand this - `optionsCorrect` is true if `options` is an array *or* has a length < 1?

src/components/TextField.js
```diff
@@ -85,6 +100,15 @@ class TextFieldComponent extends PureComponent {
...
+        {this.props.options && !optionsCorrect && (
+          <ErrorText
+            show={this.props.options}
+            className={'smc-text-field-error-text'}
+            errorTextStyle={this.props.errorTextStyle}
+          >
+            Must have an array of at least one option passed in
+          </ErrorText>
+        )}
         <ErrorText
```
martahj: Might want to have some kind of check so that only one ErrorText can render at once

src/components/TextField.js
ceciliamvrie: My logic was pretty wrong there, lol. Thanks for pointing that out!

src/components/TextField.js
ceciliamvrie: I was thinking that I could add this in for the optionsError
```
show={!hasError}
```

martahj: That could work, or you could pass the !optionsCorrect error into the one already rendering - either approach would be good!

src/components/DropdownMenu.js
```diff
@@ -0,0 +1,128 @@
...
+
+export default class DropdownMenu extends Component {
+  state = {
+    isOpen: false,
+    options: this.props.options || undefined,
+    selected: this.props.options ? this.props.options[0] : undefined,
+    isChrome: undefined,
+  }
+
+  componentWillMount = () => {
```
martahj: [Q] What's the reason for putting this logic in `componentWillMount` rather than in the constructor?

src/components/DropdownMenu.js
```diff
@@ -0,0 +1,128 @@
...
+  }
+
+  componentDidMount = () => {
+    if (navigator) {
+      const isChrome = Boolean(/Chrome/.test(navigator.userAgent) && /Google Inc/.test(navigator.vendor));
+      this.setState({ isChrome });
+    }
+  }
+
+  updateStateToDefaultOption = (defaultOptionIndex) => {
```
martahj: [Q] It looks like this function is only called in the initial state setup, what's the reason for adding it as a class method?

src/components/DropdownMenu.js
```diff
@@ -0,0 +1,128 @@
...
+  }
+   
+  onSelectMenuItem = (option): void => {
+    this.setState({ selected: option });
+  }
+
+  toggleSelect = (e): boolean => {
+    e.preventDefault();
+    this.setState({ isOpen: !this.state.isOpen });
+    return false;
```
martahj: [Q] What happens if you don't return false?

src/components/TextField.js
```diff
@@ -45,6 +46,13 @@ class TextFieldComponent extends PureComponent {
   };
 
   render() {
+    const hasValidOptions = this.props.options && (
+      Array.isArray(this.props.options) &&
+      this.props.options.length > 0
+    );
+    const hasValidDefault = (this.props.options === undefined && this.props.defaultOption) ||
+    (hasValidOptions && this.props.defaultOption);
```
martahj: [Q] What would be the use case for passing in a default option but no list of options?

src/components/DropdownMenu.js
ceciliamvrie: That was originally there as support for the Safari browser, in case e.preventDefault doesn't work

src/components/TextField.js
ceciliamvrie: I believe @danielstclair wanted support for no options being passed in, but a defaultOption gets passed in

src/components/DropdownMenu.js
ceciliamvrie: I put it here because it handles a lot of Boolean logic for the props that are being passed in, and I figured it would be way too much to fit into the constructor

pages/text-fields.js
```diff
@@ -114,6 +114,28 @@ class TextFieldPage extends PureComponent {
...
+          <TextFieldWithBottomMargin
+              floatingLabelText="Dropdown Menu"
+              options={["option1", "option2"]}
+              defaultOption={"select an option"}
+          />
+          <h4>Selected Default Option</h4>
+          <TextFieldWithBottomMargin
+              floatingLabelText="Dropdown Menu"
+              options={["option0", "option1", "option2"]}
+              defaultOption={"option2"}
```
danielstclair: [q] Why did we get rid of the dropdown prop? It seemed more explicit. that way

src/components/DropdownMenu.js
ashwoodall: [Q] Does this need to be an arrow function?

src/components/TextField.js
danielstclair: I think that checking for the defaultOptions and the options array would be better suited inside the Dropdown component.

pages/text-fields.js
ceciliamvrie: @martahj suggested that it would take care of  eliminating the problem with the passing in undefined options as props

src/components/DropdownMenu.js
```diff
@@ -0,0 +1,128 @@
...
+  bottom: 5px;
+  fill: #726969;
+`;
+
+
+export default class DropdownMenu extends Component {
+  state = {
+    isOpen: false,
+    options: this.props.options || undefined,
+    selected: this.props.options ? this.props.options[0] : undefined,
```
danielstclair: Two two things:

I think the options key should be an array if we don't pass it in so we don't get an error when trying to map over it in the render.

The selected should be this.props.defaultValue || '' and then we can be clearer on how something becomes the value initially

src/components/DropdownMenu.js
ceciliamvrie: @ashwoodall I guess not, I just wanted it to be inline styles-wize with everything else

src/components/TextField.js
ceciliamvrie: @danielstclair It was intended to be there to display to the user if they pass in the wrong options

pages/text-fields.js
martahj: lol yeah that was my idea. I thought that we could just use whether or not options was an array to determine if the intent was to create a dropdown because no other textfields use options

src/components/DropdownMenu.js
ceciliamvrie: 👍

src/components/TextField.js
ceciliamvrie: 👍

src/components/TextField.js
martahj: @ceciliamvrie What are you thinking of doing for this?

ceciliamvrie: @martahj it's in case if the user passes in invalid options. Either the options array is empty, or it's not an array

martahj: Ok I see, you added `show={!hasError}` 👍

martahj:
@ashwoodall Could you take another pass at this PR when you get a chance?

ceciliamvrie:
@danielstclair Sounds great!

PR #168, opened 2018-03-03
feat(Table): Checkboxes + Sorting!

BennettStaley:
Added checkboxes and filtering to the table component. See the new notes on the deployment below!

Very excited to share this. Tear it up! 

(be sure to check the tables page below for usage of the new props!)

https://docs-ibzdgzpvwa.now.sh

### DIFF
--- a/pages/tables.js
+++ b/pages/tables.js
@@ -13,7 +13,6 @@ const fields = [
     key: 'event',
     label: 'Event',
     numerical: false,
-    sortable: true,
   },
   {
     key: 'price',
@@ -190,7 +189,7 @@ const Tables = () => (
               it is passed the Datum of that row as its first param
             </li>
             <li>
-              onUnheck: () => any
+              onUncheck: () => any
               Callback that will be called when the user clicks a checkbox off.
               it is passed the Datum of that row as its first param
             </li>
--- a/pages/tables.js
+++ b/pages/tables.js
@@ -1,4 +1,4 @@
-import React, { Component } from 'react';
+import React, { PureComponent } from 'react';
 import MaterialThemeProvider from '../src/theme/ThemeProvider';
 import { Table } from '../src/components/Table';
 
@@ -7,16 +7,19 @@ const fields = [
     key: 'date',
     label: 'Date',
     numerical: false,
+    sortable: true,
   },
   {
     key: 'event',
     label: 'Event',
     numerical: false,
+    sortable: true,
   },
   {
     key: 'price',
     label: 'Price',
     numerical: true,
+    sortable: true,
   },
 ];
 
@@ -26,6 +29,24 @@ const data = [
     date: '02/01/17',
     event: 'Price Change',
     price: '$24,500,000',
+  },
+  {
+    key: 'xzvxzcv',
+    date: '02/01/17',
+    event: 'Open House',
+    price: '$5,500,000',
+  },
+  {
+    key: 'ssssdas',
+    date: '02/01/17',
+    event: 'Showing',
+    price: '$14,500,000',
+  },
+  {
+    key: 'efsadf',
+    date: '02/01/17',
+    event: 'Price Change',
+    price: '$774,500,000',
   },
   {
     key: '2398ro829j',
@@ -63,7 +84,7 @@ const StyledTable = Table.extend`
 const incrementCurrentPage = ({ currentPage }) => ({ currentPage: currentPage + 1 });
 const decrementCurrentPage = ({ currentPage }) => ({ currentPage: currentPage - 1 });
 
-class ControlledTable extends Component {
+class ControlledTable extends PureComponent {
   state = {
     currentPage: 1,
   };
@@ -116,6 +137,9 @@ const Tables = () => (
               <li>
                 numerical?: boolean
               </li>
+              <li>
+                sortable?: boolean
+              </li>
             </ul>
           </li>
           <li>
@@ -153,6 +177,41 @@ const Tables = () => (
             handleBackwardsPagination: () => any
             Callback that will be called when the user clicks the 'back' arrow
           </li>
+          <br />
+          <li>
+            hasCheckboxes?: boolean
+            denotes checkboxes on items, if this prop is passed you must also pass the
+            next two methods below, for onCheck and onUncheck
+          </li>
+          <ul>
+            <li>
+              onCheck: () => any
+              Callback that will be called when the user clicks a checkbox on.
+              it is passed the Datum of that row as its first param
+            </li>
+            <li>
+              onUnheck: () => any
+              Callback that will be called when the user clicks a checkbox off.
+              it is passed the Datum of that row as its first param
+            </li>
+          </ul>
+          <br />
+          <li>
+            notes on sorting:
+          </li>
+          <ul>
+            <li>
+              if Uncontrolled, the sort will order all the data regardles of the currentPage
+              this is because it manipulates all the data it was given. Page agnostic.
+            </li>
+            <li>
+              if controlled via a parent, it will manipulate only the data on that page.
+              as that is all it was given.
+            </li>
+            <li>
+              The sorting method is called NaturalSort. Credits to: <a href='https://github.com/bubkoo/natsort' target='none'>https://github.com/bubkoo/natsort</a>
+            </li>
+          </ul>
         </ul>
       </h4>
       <h2>Regular Table</h2>
@@ -176,8 +235,35 @@ const Tables = () => (
       />
       <h2>Controlled table with footer</h2>
       <ControlledTable />
+      <h2>Table checkboxes</h2>
+      <Table
+        hasCheckboxes
+        onCheck={() => {
+          console.log('select');
+        }}
+        onUncheck={() => {
+          console.log('deselect');
+        }}
+        fields={fields}
+        data={data}
+        header={`Check the console for callbacks!`}
+      />
+      <h2>Table checkboxes, with sorting!</h2>
+      <Table
+        hasCheckboxes
+        rowsPerPage={2}
+        onCheck={(e) => {
+          console.log('select', e.key);
+        }}
+        onUncheck={(e) => {
+          console.log('deselect', e.key);
+        }}
+        fields={fields}
+        data={data}
+        header={`Check the console for callbacks!`}
+      />
     </div>
-  </MaterialThemeProvider>
+  </MaterialThemeProvider >
 );
 
 export default Tables;
--- a/src/components/Table/Table.js
+++ b/src/components/Table/Table.js
@@ -1,5 +1,7 @@
 import React, { PureComponent } from 'react';
 import styled from 'styled-components';
+import isEqual from 'lodash.isequal';
+
 import Row from './Row';
 import Datum from './Datum';
 import Title from './Title';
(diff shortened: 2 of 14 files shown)


pages/tables.js
```diff
@@ -153,6 +177,41 @@ const Tables = () => (
...
+            next two methods below, for onCheck and onUncheck
+          </li>
+          <ul>
+            <li>
+              onCheck: () => any
+              Callback that will be called when the user clicks a checkbox on.
+              it is passed the Datum of that row as its first param
+            </li>
+            <li>
+              onUnheck: () => any
```
ashwoodall: heckin Unheck

BennettStaley:
For now, when pagination is in place, we disable the checkbox. This is because changing the checkbox's checked prop triggers the onChange function, unchecking the previous page's items.

src/components/Table/Table.js
```diff
@@ -1,35 +1,167 @@
...
 
-const incrementCurrentPage = ({ currentPage }) => ({ currentPage: currentPage + 1 });
-const decrementCurrentPage = ({ currentPage }) => ({ currentPage: currentPage - 1 });
 
-class Table extends Component {
-  state = {
-    currentPage: this.props.currentPage || 1,
+const UpwardArrow = props => <svg {...props} viewBox="0 0 24 24">
+  <path d="M4 12l1.41 1.41L11 7.83V20h2V7.83l5.58 5.59L20 12l-8-8-8 8z" />
+</svg>;
```
martahj: [Q] Is this a material icon?

src/components/Table/Table.js
```diff
@@ -1,35 +1,167 @@
...
+      sortedBy: '',
+      descending: false,
+      sortedData: this.props.data,
+      currentPage: this.props.currentPage || 1,
+    };
+    this.rowsPerPage = this.props.rowsPerPage || 10;
+  }
+
+  componentWillReceiveProps(nextProps) {
+    if (this.props.data !== nextProps.data) {
```
martahj: This will do a diff based on reference equality rather than value equality. Is that the intended behavior?

src/components/Table/Table.js
```diff
@@ -133,10 +287,54 @@ class Table extends Component {
...
+        fill: rgba(0, 0, 0, .54);
+        transform-origin: center;
+        transition: 0.3s;
+      }
+
+      > .sortButton:hover {
+        background-color: rgba(0, 0, 0, .04);
+      }
+
+      > .rotate {
```
martahj: The rotation looks really good

src/components/Table/Table.js
BennettStaley: It works to handle when the pagination is controlled by the parent. it should be fine doing a reference diff, right?

src/components/Table/Table.js
BennettStaley: love me some `transition: 0.2s`!

src/components/Table/Table.js
BennettStaley: Talked offline with @martahj, added in additional checks!

src/icons/icons.js
```diff
@@ -1,6 +1,7 @@
 import * as React from 'react';
 import styled from 'styled-components';
 import { ArrowDropDownPath } from './ArrowDropDownPath';
+import { UpwardArrowPath } from './UpwardArrowPath';
```
martahj: [REQ] Since we're ultimately going to have a lot of icons, it seems important that we set the names to exactly what they are in material so users have no problem finding them. To that end, could you rename UpwardArrow to [ArrowUpward](https://material.io/icons/#ic_arrow_upward)?

screen shot 2018-03-05 at 11 06 37 am

src/components/Table/Table.js
```diff
@@ -1,35 +1,175 @@
...
+    };
+
+    if (this.props.data !== nextProps.data) {
+      setStort();
+    } else if (this.props.data.length !== nextProps.data.length) {
+      setStort();
+    } else if (!isEqual(this.props.data, nextProps.data)) {
+      setStort();
+    } else {
+      this.setState({ sortedData: nextProps.data });
```
martahj: Is there a reason for setting sortedData to the new data if we believe it hasn't changed?

BennettStaley:
@martahj changes made!

brad-decker:
@BennettStaley -- is the link in the description the most up to date?

src/components/Table/Table.js
```diff
@@ -1,35 +1,175 @@
...
+      this.setState({ sortedData: data });
+    };
+
+    if (this.props.data !== nextProps.data) {
+      setStort();
+    } else if (this.props.data.length !== nextProps.data.length) {
+      setStort();
+    } else if (!isEqual(this.props.data, nextProps.data)) {
+      setStort();
+    }
```
martahj: Not a change request, but since these all these if statements have the same result you could combine them into one large boolean check

src/components/Table/Table.js
```diff
@@ -1,35 +1,175 @@
...
+      this.props.onUncheck && this.props.onUncheck(item);
+      delete itemsToSet[item.key];
+    }
+    this.setState({
+      selectedItems: itemsToSet,
+      selectedAllActive: false,
+    });
+  }
+
+  sortBy(key) {
```
martahj: This isn't really necessary for this PR to land, but a future enhancement might be to allow the sorting to be controlled by the parent, so the parent could eg issue a relay query in response to an interaction with the table.

src/components/Table/Table.js
```diff
@@ -1,35 +1,175 @@
...
+    let shouldDescend;
+    if (key === this.state.sortedBy) {
+      // flip descending or ascending
+      if (this.state.descending) {
+        shouldDescend = false;
+      } else {
+        shouldDescend = true;
+      }
+    } else {
+      // default, new stort to descending
```
martahj: stort? 😛

src/components/Table/Table.js
BennettStaley: yes! i plan to add that in a future pr alongside filtering (searching)

src/components/Table/Table.js
BennettStaley: lmao

BennettStaley:
@petegivens thanks for the heads up! I'm new to using the theme prop. made those changes and a few more!

@brad-decker yes it is, i update it with all changes made.

brad-decker:
@BennettStaley sending a screen recording of something weird that I witnessed while messing around with all the tables.  https://drive.google.com/file/d/1RWDy1nUlIWoj4cXZmzc_VS7a4R90NGmY/view?usp=sharing

BennettStaley:
great catch @brad-decker I was literally _just_ looking into this.

It appears to be a data mutation case. One table is mutating the constants defined on the page and the others are accessing it.

martahj:
@BennettStaley Have you fixed the issue that @brad-decker found?

brad-decker:
Appears so.

src/components/Table/Table.js
```diff
@@ -1,35 +1,179 @@
...
+    super(props);
+    this.state = {
+      selectedAllActive: false,
+      selectedItems: {},
+      sortedBy: '',
+      descending: false,
+      sortedData: this.props.data,
+      currentPage: this.props.currentPage || 1,
+    };
+    this.rowsPerPage = this.props.rowsPerPage || 10;
```
brad-decker: [Q] Why is this on the instance of the component and not in state?

src/components/Table/Table.js
```diff
@@ -1,35 +1,179 @@
...
+  currentPage: currentPage + 1,
+});
+const decrementCurrentPage = ({ currentPage }) => ({
+  currentPage: currentPage - 1,
+});
+
+class Table extends PureComponent {
+  constructor(props) {
+    super(props);
+    this.state = {
```
brad-decker: [Q]: looks like we moved away from using state as a class property and moved it into the constructor, can you clarify the reasoning behind this?

src/components/Table/Table.js
```diff
@@ -1,35 +1,179 @@
...
+    // first we see if the ref is different
+    // then we see if the length is different
+    // then we deep diff the two
+
+    // this is to catch weird usage of the comp
+    // ie: pushing or popping from an array
+
+    const setSort = () => {
+      let sortedData;
+      const currentSort = this.state.sortedBy;
```
brad-decker: [Q]: any potential negative ramifications of basing the computations in this method from current state and then calling setState later? I realize you're modifying a different state key so probably not a issue but curious to your thoughts on this? Would it ever be possible for sortedBy to change before this method completes causing weird behavior?

brad-decker: [Q addendum]: Oh snap, i was confused because I didn't realize this was a part of componentWillReceiveProps. setStates are batched inside of lifecycle methods. Could we extract this method above the class scope, have it accept arguments for state and props and return the new state tree? This will prevent us from creating a function everytime the component's props update

src/components/Table/Table.js
BennettStaley: Good question, unfortunately i didn't write that, I simply moved it into the constructor. My guess is that it doesn't need to trigger a re-render when it changes (or it never changes and is just instantiated) so its not necessary to be in state. Just a guess. see line 32 on the unchanged Table.js

src/components/Table/Table.js
BennettStaley: Sure, I think thats a great idea.

src/components/Table/Table.js
BennettStaley: Could you talk about the pros and cons of how either methodology would affect this component? I figured that setting our state to the props we are given would be a one time instantiation.

src/components/Table/Table.js
brad-decker: Good answer.

src/components/Table/Table.js
brad-decker: So its more of a preference versus a pro vs con of performance. With ES7 class properties *eventually* they will have a performance boost over straight constructor methods. The reason being is that they are instantiated a lower level in the creation chain. https://github.com/tc39/proposal-class-fields -- also eventually we will get support for private fields allowing us to hide properties completely. These all require changes to the core of javascript to work. 

Presently the way the babel class-properties transform works is it just moves everything into the constructor -- so function wise there is no difference. Lets take a look at an example outside of React.Component:

```js
class Employee {
  title = 'sensible_default';
  salary = 0.00;
  constructor(title, salary) {
    this.title = title;
    this.salary = salary;
  }
```

^- in this example both class properties and constructor are necessary. The class properties set the sensible default. Now lets extend it. 

```js
class EmployeeOfMerlinLabs extends Employee {
  salary = this.salary * 1.25;
  title = `Wizard of ${this.title}`;
}
```

^- here the constructor is already called from Employee so this.title and this.salary are already set. This is the same case with react.Component. this.props will be provided. 

Its not necessary to change this part of your PR right now -- the benefits of using class properties are not very high right now. But in the future our understanding of constructors will change. In many cases we will utilize both class properties and constructors. Class properties help to identify whether the property is private or public and also, with flow, the type that it should be.

brad-decker:
Also just in case it wasn't stated before: This is awesome work @BennettStaley.

src/components/Table/Table.js
martahj: It will be moved into state at some point. I was using that as a convention to make it clearer which aspects could change and which could not in the initial implementation

PR #169, opened 2018-03-05
feat(Checkbox): improve handling of indeterminate

petegivens:
API change: `defaultChecked` prop removed and replaced with a `default` prop which can accept 'checked' or 'indeterminate'. 

Should allow forms that are controlling checkboxes to reset them back to an indeterminate state.


petegivens:
@BennettStaley this should help if you need your lists/forms to be able to reset a checkbox to its indeterminate state.

PR #171, opened 2018-03-05
feat(List): Add ListSubheader component to List

hsarb:
Added a ListSubheader to the List component. Supports default and primary color rules.

I had to add an `index.js` file to the menu component as the build was unable to resolve the menu path.

https://docs-wsuneqahya.now.sh

PR #173, opened 2018-03-05
feat(List): Add ListSubheader to List component

hsarb:
Adding support for Subheaders in the List component. Supports both default and primary color options by passing in a primary prop.

Does not currently support alignment for the case when a floating action button is left aligned.

https://docs-bscxeohlhj.now.sh


hsarb:
@brad-decker my habit of merging vs rebasing caused some funkiness with #171. Reopening here to keep it clean

PR #178, opened 2018-03-06
feat(Table): Add Search

BennettStaley:
Adds Search functionality to the table. This comes in the form of a styled search that fires an onSearch prop that the user provides!

Wanted to get feedback on how this guy should work. I'm having a hard time finding design specs for material tables with search components. therefore I decided to make this as best i could.

https://docs-jdfbczswgo.now.sh/tables/

### DIFF
--- a/src/components/Table/Search.js
+++ b/src/components/Table/Search.js
@@ -1,0 +1,45 @@
+import React from 'react';
+import styled from 'styled-components';
+import TextField from '../TextField';
+import { SearchIcon } from '../../icons/icons';
+
+const SearchComponent = ({ className, onSearch }) => (
+  <div className={`smc-search-container ${className}`}>
+    <div className={`smc-search-icon-container`}>
+      <SearchIcon fill='black' />
+    </div>
+    <TextField
+      type='search'
+      hintText={'Search Term'}
+      fullWidth
+      onChange={onSearch}
+    />
+  </div>
+);
+
+const Search = styled(SearchComponent) `
+  padding: 25px 30px 20px 30px;
+  width: 100%;
+  margin: auto;
+  display: inline-flex;
+  background-color: white;
+  box-sizing: border-box;
+  position: relative;
+
+  ${TextField} {
+    margin: auto;
+    margin-left: 15px;
+    height: 28px;
+    border-bottom: none;
+    position: relative;
+  }
+
+  ${SearchIcon} {
+    position: relative;
+    bottom: 0px;
+    margin: auto;
+    height: 28px;
+  }
+`;
+
+export default Search;
--- a/src/components/Table/Table.js
+++ b/src/components/Table/Table.js
@@ -8,6 +8,7 @@ import Title from './Title';
 import Header from './Header';
 import Footer from './Footer';
 import naturalSort from './naturalSort';
+import Search from './Search';
 import Checkbox from '../Checkbox';
 import { ArrowUpwardIcon } from '../../icons/icons';
 
@@ -40,7 +41,7 @@ class Table extends PureComponent {
       selectedItems: {},
       sortedBy: '',
       descending: false,
-      sortedData: this.props.data,
+      mutatedData: [...this.props.data],
       currentPage: this.props.currentPage || 1,
     };
     this.rowsPerPage = this.props.rowsPerPage || 10;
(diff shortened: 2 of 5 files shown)


BennettStaley:
@hsarb Great question, that's exactly why I opened this up for discussion! I've been unable to find any Material spec outlining anything related to table search and have since decided to start looking at other open source repos for inspiration.

BennettStaley:
Yes. We had a bit of a debate on which to settle on, but you're totally right. I realize this pr would lead to a fork of what had originally been planned. I'm gonna tag @martahj and @brad-decker to chime in on this.

danielstclair:
All great points here. I think that the search-inside-of-table functionality should be something left outside of this library. Being that a search is a textfield I'd rather keep them separate in the library and allow the user to compose them together.

https://material.io/guidelines/components/text-fields.html#text-fields-field-variations
https://material.io/guidelines/patterns/search.html

brad-decker:
@hsarb @BennettStaley I'm not opposed to adding new components that are inspired by the Material Spec but in general functionality should be left to userland. For example sorting and searching are things that can't be done effectively straight from a library. We won't be able to use a sort feature or search feature that is baked into a library nor would we be using pagination. We'd want to use Relay for these things and do sorting, search and what not via our providers or our own api.

Its fine to create hooks for those things so that the table behaves as it should in those situations but it shouldn't actually try to sort or search on the table.

hsarb:
Agreed!

brad-decker:
So I spoke with @BennettStaley and got some additional clarity. He is not envisioning implementing the logic for search in the library but rather giving the user an option to add a search feature to the table which would provide hooks for the user to implement the search logic. To be clear the material spec is about components and about a guideline. If you put a material search bar into a material table it is still material spec and I think more customizable for the end user. It should obviously not enforce a search bar but give the user the option to display one.

src/components/Table/Table.js
```diff
@@ -39,8 +40,9 @@ class Table extends PureComponent {
       selectedAllActive: false,
       selectedItems: {},
       sortedBy: '',
+      filteredBy: 'Price Change',
```
martahj: [Q] What’s the reason for setting this as the default value on the table? Seems like most data sets wouldn’t have this

src/components/Table/Table.js
```diff
@@ -155,6 +156,10 @@ class Table extends PureComponent {
     });
   }
 
+  filterBy = () => {
```
martahj: [Q] I’m having trouble finding where this is used for some reason, could you point me to the line lol? Ctl F on mobile may be bad

src/components/Table/Search.js
```diff
@@ -0,0 +1,45 @@
...
+const Search = styled(SearchComponent) `
+  padding: 25px 30px 20px 30px;
+  width: 100%;
+  margin: auto;
+  display: inline-flex;
+  background-color: white;
+  box-sizing: border-box;
+  position: relative;
+
+  > .smc-text-field-container {
```
brad-decker: could we import the text field container component and use the ${TextFieldContainer}? here

src/components/Table/Search.js
```diff
@@ -0,0 +1,45 @@
...
+    margin: auto;
+    margin-left: 15px;
+    height: 28px;
+    border-bottom: none;
+    position: relative;
+  }
+
+  > .smc-search-icon-container {
```
brad-decker: Same here

src/components/Table/Table.js
```diff
@@ -172,11 +172,10 @@ class Table extends PureComponent {
     const sorter = naturalSort({ desc: shouldDescend });
 
     // sort by key!
-    const data = [...this.state.sortedData];
-    const sorted = data.sort(
+    const sorted = this.state.mutatedData.sort(
```
martahj: I'm not sure if it'll make a difference given that the variable is called *mutated*Data, but just to confirm - `.sort` will mutate the original array, is that the intended/expected behavior? 

```js
const a = [2, 1, 3]
const b = a.sort()
console.log(a) // [1, 2, 3]
```

BennettStaley: yes. The .sort method will mutate the array. I've made precautions to always [...data] when needed. (interacting with props)

BennettStaley:
updated the link!

src/components/Table/Table.js
```diff
@@ -100,46 +101,46 @@ class Table extends PureComponent {
...
       selectedItems: itemsToSet,
       selectedAllActive: selectedAll,
     });
   }
 
 
-  updateWithNewProps = (newData) => {
-    let sortedData;
+  updateWithNewProps = (input) => {
+    const newData = [...input];
```
AriLFrankel: not a change request, but I think this could be simplified by removing newData

```
const mutatedData = [...input];
...
if(currentSort){
...
mutatedData.sort(...)
}
this.setState({mutatedData})
```

BennettStaley: Noted. I'll update it in my next pr.

src/components/Table/Table.js
```diff
@@ -313,6 +316,13 @@ export default styled(Table) `
   border-spacing: 0;
   border: 0px;
 
+  > .smc-table-search {
+    // not sure if we should include a border top, looks odd to me
+    // thoughts?
```
AriLFrankel: To me it seems like we could leave it up to the consumer

PR #186, opened 2018-03-08
refactor(Table): Remove elevation

dekkofilms:
I feel as though we should remove elevation from being on the tables. According to spec, it seems as though the table itself doesn't have elevation on it, but rather inserting the tables into cards.

Right now we have tables within cards, and it now has double elevation.

Here is the spec, with the section I'm referring to being (Tables within cards): https://material.io/guidelines/components/data-tables.html#data-tables-tables-within-cards

Now Link: https://docs-sndqbfdqsj.now.sh

### DIFF
--- a/src/components/Table/Table.js
+++ b/src/components/Table/Table.js
@@ -8,7 +8,6 @@ import Title from './Title';
 import Header from './Header';
 import Footer from './Footer';
 import naturalSort from './naturalSort';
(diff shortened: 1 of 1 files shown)


BennettStaley:
yeuuup. I left it in because it looked _cool_ but you're right, it should be left up to the user to add this.

PR #192, opened 2018-03-09
fix(Tables): Handle Empty Datum

BennettStaley:
Return null for datum with no children
(see faked pagination for an example!)

Fix checkbox header float left on fullWidth

https://docs-icojuhrnkl.now.sh/tables/

### DIFF
--- a/pages/tables.js
+++ b/pages/tables.js
@@ -64,16 +64,9 @@ const data = [
     event: 'Something else',
     price: '$34,232,544',
   },
-  {
-    key: '4534r34',
-    date: '03/14/15',
-    event: 'Really Big Price Change',
-    price: '$4,000',
-  },
 ];
 
 const StyledTable = Table.extend`
-  max-width: 500px;
   .smc-table-datum-price {
     color: red;
   }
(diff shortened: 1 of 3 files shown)


pages/tables.js
```diff
@@ -64,16 +64,16 @@ const data = [
...
+  //   key: '4534r34',
+  //   date: '03/14/15',
+  //   event: 'Really Big Price Change',
+  //   price: '$4,000',
+  // },
 ];
 
 const StyledTable = Table.extend`
-  max-width: 500px;
+  // max-width: 500px;
```
martahj: [Q] This too

pages/tables.js
```diff
@@ -64,16 +64,16 @@ const data = [
...
-    date: '03/14/15',
-    event: 'Really Big Price Change',
-    price: '$4,000',
-  },
+  // {
+  //   key: '4534r34',
+  //   date: '03/14/15',
+  //   event: 'Really Big Price Change',
+  //   price: '$4,000',
+  // },
```
martahj: [Q] Any reason this is commented out rather than deleted?

BennettStaley: This is commenting out the data so that we have an odd number of fields. I'll just delete it instead.

BennettStaley:
⚠️  link uuuupdated!

Issue #193, opened 2018-03-09
Snackbar/Toasts do not show up on safari


brad-decker:
In our project we are noticing that snackbars aren't showing up the same way in safari as they are in chrome. This issue is meant to keep track of that and to report findings. It could be an implementation issue or it could be a SMC issue.

PR #194, opened 2018-03-09
fix(Snackbar): Remove from DOM if closed

dekkofilms:
This is setting the Snackbar to display none if it is not yet opened. In the Merlin repo it's easy to see the error / recreate it, but in SMC it's hard to reproduce the same effect.

I'm hoping this could potentially resolve what is happening in this issue #193.


martahj:
@dekkofilms Could you post a now link?

dekkofilms:
Yeah! Sorry for the delay @martahj 

https://docs-emdihozxtq.now.sh

PR #200, opened 2018-03-13
feat(Tabs): Add animations and general cleanup

hsarb:
Added the following features:

- Inkbar Animation
- Ripple
- Icon layout
- Fixed layout

Also cleaned up the props and code a bit. There is an issue with the ripple randomly not working, not sure if this is a known issue with the ripple enhancer.

https://docs-nkrlmoijni.now.sh/tabs/


src/components/Tabs/InkBar.js
```diff
@@ -1,22 +1,18 @@
 import React from 'react';
 import styled from 'styled-components';
 
-const InkBarContainer = styled.div`
-  position: relative;
-  bottom: 0;
-  left: ${props => props.selectedIndex * props.tabWidth}px;
-  width: ${props => props.tabWidth}px;
-  height: 2px;
+const InkBarComponent = ({ className }) => <div className={`${className} smc-tab-bar-inkbar`} />;
```
brad-decker: instead of using component wrappers and applying classNames, the new API we are trying to achieve is to use something like this:
```
& ${ComponentName} {
  ...styles...
}
```

it allows us to get rid of these intermediary components, too:

```js
styled.div.attrs({
  'data-smc': 'InkBar'
})`
...inkbarStyles...
`;
```

src/components/Tabs/Tab.js
```diff
@@ -1,63 +1,54 @@
...
-  font-size: 0.875rem;
-  font-weight: 500;
-  letter-spacing: 0.04em;
-  line-height: 1.5rem;
-  display: table-cell;
+const RippleTab = withRipple(StyledTab);
+
+const TabComponent = ({ className, fontColor, icon, label, onClick, selected, index }) => (
+  <RippleTab
+    className={`${className} smc-tab`}
```
brad-decker: lets drop the custom 'smc-tab' classname in favor of data-smc="Tab"

src/components/Tabs/Tab.js
```diff
@@ -1,63 +1,54 @@
...
+    onClick={() => onClick(index, label)}
+    selected={selected}
+    fontColor={fontColor}
+  >
+    {icon}
+    {label && <span className="smc-tab-label">{label}</span>}
```
brad-decker: lets create a styled.span for this label, with a data-smc prop of 'TabLabel'

src/components/Tabs/Tab.js
```diff
@@ -1,63 +1,54 @@
...
-    {label}
-  </TabContainer>
-);
+  > svg {
+    fill: ${props => props.fontColor || props.theme.white};
+  }
 
-const Tab = styled(TabComponent)``;
+  > .smc-tab-label {
+    padding: 4px 16px;
```
brad-decker: Move these styles into the new TabLabel styled.span

src/components/Tabs/Tabs.js
```diff
@@ -1,115 +1,111 @@
...
   }
 }
 
-const Tabs = styled(TabsComponent)``;
+const Tabs = styled(TabsComponent)`
+  overflow: hidden;
+  position: relative;
+  width: 100%;
+
+  > .smc-tab-navigation {
```
brad-decker: I'd like to also see this moved to a separate component that is exported (along with the others I've suggested) for the purposes of styling subcomponents.

src/components/Tabs/InkBar.js
hsarb: Ok, trying to wrap my head around this in the context of InkBar. As a single element how should this be done? I understand how to with multiple elements

hsarb: nm, figured it out

hsarb:
@brad-decker updated to follow the new API Style. Let me know if thats what you envisioned.

brad-decker:
This looks great! Thanks for refactoring to match the intended API!

PR #202, opened 2018-03-13
feat(Table): Check all + Direct Control

BennettStaley:
Assume direct control over the table!

Check all per pages! WOW!


BennettStaley:
We now do a diff to see if we have a specific page checked. 
```
    const currentSelectionKeys = Object.keys(selectedItems);
    const currentPageKeys = rowsToShow.map(item => item.key);
    const selectedAllActive = this.determineSelected(currentSelectionKeys, currentPageKeys);
```

this happens in the render. Our superSet is the currentSelection keys, an array of all the currently selected objects' keys. And our subSet is all the keys on the current page. That way we can tell if all of the current page keys are contained within the selected keys, and thus change the state of the header checkbox.

This also changes where the on-click happens for the header checkbox, moving it to the parent container. That way when we un-check an item via the prop, we aren't calling the uncheck all function unless it was a real click.

BennettStaley:
Additionally we now can control everything about the table! Even sorting!
```
        onSearch
        onSort
        sortedBy
        hasCheckboxes
        onCheck
        onUncheck
        descending
        rowsPerPage
        handleForwardPagination
        handleBackwardsPagination
```
See the tables page for more info on these props!

PR #205, opened 2018-03-14
feat(Table): RowsPerPage

BennettStaley:
Implement rows per page drop down!

You can now pass a rowsDropDown optional prop to control what selection options there are. if you pass an empty array it wont render!

https://docs-znlmpgbjwm.now.sh/tables/ !!!


petegivens:
Seems like there might be a bug that causes the Rows Per Page dropdown to disappear when a large number of rows is selected. Here's a GIF: https://drive.google.com/file/d/1T9QEFVzJxS9nIyGsR9hVqja8Yi9pciWW/view?usp=sharing

Is that intended behavior?

BennettStaley:
@petegivens, yeah i noticed that. The other option is to always have the footer be there. Thoughts on how it _should_ work?

petegivens:
@BennettStaley IMO the footer should persist. Why is it disappearing?

BennettStaley:
it WAS because the footer wasn't showing if all the data was going to fit. Now if you check the demo it should persist. @petegivens

src/components/Table/Footer.js
```diff
@@ -1,13 +1,23 @@
-import React from 'react';
+import React, { Fragment, PureComponent } from 'react';
 import styled from 'styled-components';
 import Button from '../Button';
 import { Icon } from '../../icons/icons';
+import { Menu } from '../Menu';
 
 const FooterIcon = Icon.extend.attrs({
+  cursor: 'pointer',
   fill: 'rgba(0, 0, 0, .54)',
```
petegivens: [Q] Does it make sense to lock the icons into using the same color as primary text? If so, we can get this color from `props.theme.textColors.primary`. If not, we may want to think about adding an `icon` color prop to the theme.

petegivens: [Q] Does it make sense to lock the icons into using the same color as primary text? If so, we can get this color from `props.theme.textColors.primary`. If not, we may want to think about adding an `icon` color prop to the theme.

PR #207, opened 2018-03-15
feat(Textarea): Add textarea with border

### LINKED ISSUE #190: TextField component's FloatingLabel should be a label element
The internal FloatingLabel component for a TextField should be refactored to be a label component for better a11y. This would also require a refactor of the FloatingLabel CSS as labels and divs interact differently.

### LINKED ISSUE #198: Textarea border
Textareas should support borders

https://material.io/guidelines/components/text-fields.html#text-fields-field-types

ashwoodall:
Add text area with border

https://material.io/guidelines/components/text-fields.html#text-fields-field-types

Demo: https://docs-scyqspdjkm.now.sh/text-fields/

Fixes #198 
Fixes #190

### DIFF
--- a/pages/text-fields.js
+++ b/pages/text-fields.js
@@ -125,6 +125,12 @@ class TextFieldPage extends PureComponent {
             rows={2}
             charLimit={100}
           />
+          <TextFieldWithBottomMargin
+            textarea
+            hasBorder
+            floatingAreaLabelText="Border textarea"
+            rows={4}
+          />
           <br />
 
           <h2>Dropdown Menu</h2>
--- a/src/components/TextField.js
+++ b/src/components/TextField.js
@@ -57,6 +57,7 @@ class TextFieldComponent extends PureComponent {
         <Suffix>{this.props.suffix}</Suffix>
         <Prefix>{this.props.prefix}</Prefix>
         <FloatingLabel
+          aria-label={this.props.floatingLabelText}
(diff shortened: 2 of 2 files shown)


brad-decker:
Does this also address #190 ? @ashwoodall

ashwoodall:
@brad-decker it does now!

pages/text-fields.js
```diff
@@ -125,6 +125,12 @@ class TextFieldPage extends PureComponent {
             rows={2}
             charLimit={100}
           />
+          <TextFieldWithBottomMargin
+            textarea
+            hasBorder
+            floatingAreaLabelText="Border textarea"
```
petegivens: [Q] Is there a use case for using the regular `floatingTextLabel` with the bordered text area?

ashwoodall: floatingAreaLabelText is closer to spec for the bordered text area, but both can be applied with how the TextField component is written right now. We'll probably need to refactor this in the future to be more intuitive

petegivens: Depending on the intended use it might be fine the way it is. I was originally going to ask if you had considered eliminating the `floatingAreaTextLabel` pop and instead just inserting the right sub-component if `textarea` and `hasBorder` props were there. But then I started experimenting with adding both `floatingTextLabl` and `floatingAreaTextLabel` props and they both look good to me. Though I suppose having a floating label above a bordered text area might violate the spec

PR #208, opened 2018-03-15
feat(screensizecontext): pass platform to context

AriLFrankel:
This adds platform.js to check the user's platform and pass that
information along with the screen size context.

My thinking for including this in screen size context is that the consumers of screen size context are likely to be the ones that care about platform.

### DIFF
--- a/src/contexts/ScreenSizeContext.js
+++ b/src/contexts/ScreenSizeContext.js
@@ -1,6 +1,7 @@
 /**
  * @file ScreenSizeContext
  * @author Brad Decker <bhdecker84@gmail.com|brad@merlinlabs.com>
+ * @author Ari Frankel <ari.l.frankel@gmail.com|ari@merlinlabs.com>
  * @description Defines a context that will allow all children to
  * access the current screenSize. This is implmented using respondable.
  * All breakpoints are adjustable by the end user.
@@ -9,6 +10,7 @@ import React, { Component } from 'react';
 import createReactContext from 'create-react-context';
 import { withTheme } from 'styled-components';
 import respondable from 'respondable';
+import platform from 'platform';
 
 /**
  * Context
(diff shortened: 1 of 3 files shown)


src/contexts/ScreenSizeContext.js
```diff
@@ -34,9 +36,22 @@ export const ScreenSizeConsumer = Context.Consumer;
...
+    // Get platform data using Platform
+    // It is necessary to capture the platform data on the client.
+    // Platform relies on window.navigator.
+    // eslint-disable-next-line
+    this.setState({
+      platformData: {
+        product: platform.product,
+        osFamily: platform.os.family,
+        majorVersion: platform.version ? parseInt(platform.version.split('.')[0], 10) : -1,
+        majorOsVersion: platform.os.version ? parseInt(platform.os.version.split('.')[0], 10) : -1,
```
BennettStaley: Good call here. Some builds of linux have these numbers, some don't. Maybe add it to all the fields?

AriLFrankel: product will default to `null`
I think that there will always be an os/family. I took these configs from [concierge](https://github.com/MerlinLabs/frontend/blob/8ca7b82dd94a2973496bee3909f525916b5cb836/lib/platform.js) so I would think that would have come up in there if it were a problem not to provide a default, but couldn't hurt to add a safety margin.

brad-decker:
This is a great first pass at this implementation! I don't know if we should store these values in state when they will not change. Its not necessary to set them in state for them to work in the context just fyi. You could add additional values into the Provider value object.

Thoughts? @AriLFrankel

AriLFrankel:
@brad-decker yes I agree that they won't change. My thought in putting them there was to be able to get the platform data in didMount and then trigger a re-render to get them in to context. what are your thoughts?

brad-decker:
Beings as they are simple values we can keep this as is, however one alternative in more expensive cases is to attach to this instead of state. Beings as it only needs initialization once.

PR #214, opened 2018-03-17
feat(contexts): use HOCs

AriLFrankel:
Refactors the withRowState and withScreenSize enhancers to use HOCs passing to props in
order to support using them with class components and composition in arbitrary order.

The disadvantage of doing it this way is that we have to be more careful to avoid props clashes than with the other method. To mitigate this concern I have rowStat, screenSizeState, and platformData separated.

This is a good discussion on the subject: https://twitter.com/acdlite/status/971598256454098944

### DIFF
--- a/src/components/FlexGrid/Column.js
+++ b/src/components/FlexGrid/Column.js
@@ -1,6 +1,7 @@
 /**
  * @file Column
  * @author Brad Decker <bhdecker84@gmail.com|brad@merlinlabs.com>
+ * @author Ari Frankel <ari.l.frankel@gmail.com|ari@merlinlabs.com>
  * @description Defines a flex grid column that responds to the size
  * of the screen to grow, shrink, reorder as described by the end
  * user via size props.
@@ -209,7 +210,7 @@ const enhancer = compose(
 );
 
 const ColumnComponent = enhancer(
-  ({ className, children, ...props }, { screenSizeState, rowState }) => (
+  ({ className, children, screenSizeState, rowState, ...props }) => (
     <SizedColumn
       {...props}
       data-smc="Column"
--- a/src/components/FlexGrid/Row.js
+++ b/src/components/FlexGrid/Row.js
@@ -1,10 +1,11 @@
 /**
  * @file Row
  * @author Brad Decker <bhdecker84@gmail.com|brad@merlinlabs.com>
+ * @author Ari Frankel <ari.l.frankel@gmail.com|ari@merlinlabs.com>
  * @description Defines a flex grid row that has props defined for
  * easily setting some common flex styles.
  */
-import React, { Component } from 'react';
+import React, { Component, ComponentType } from 'react';
 import styled from 'styled-components';
 import createReactContext from 'create-react-context';
 import { rowMixin } from '../../mixins/flex';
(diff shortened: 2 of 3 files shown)


src/contexts/ScreenSizeContext.js
```diff
@@ -6,12 +6,27 @@
...
+const screenSizes = ['xs', 'sm', 'md', 'lg', 'xl', 'server'];
+export type screenSize = 'xs' | 'sm' | 'md' | 'lg' | 'xl' | 'server';
+
+export const greaterThanOrEqual = (refSize: screenSize) => (size: screenSize) =>
+  screenSizes.indexOf(size) >= screenSizes.indexOf(refSize);
+
+export const lessThanOrEqual = (refSize: screenSize) => (size: screenSize) =>
+  screenSizes.indexOf(size) <= screenSizes.indexOf(refSize);
+
+export const isDesktop = greaterThanOrEqual('lg');
```
brad-decker: What is the though behind this? Technically Desktop resolutions exist below the 'lg' key

src/contexts/ScreenSizeContext.js
```diff
@@ -6,12 +6,27 @@
...
 import createReactContext from 'create-react-context';
 import { withTheme } from 'styled-components';
 import respondable from 'respondable';
 import platform from 'platform';
 
+/*
+ * Utilities for screen size. May want to move this in to a lib file soon
+ * as it likely grows soon
+*/
+const screenSizes = ['xs', 'sm', 'md', 'lg', 'xl', 'server'];
```
brad-decker: The reason the provider uses the theme values to determine this stuff is so that the end user can override this. We should utilize the theme variables here instead of hardcoding it. You can make the methods below part of the component so they have access to the theme.

src/contexts/ScreenSizeContext.js
```diff
@@ -6,12 +6,27 @@
...
+export type screenSize = 'xs' | 'sm' | 'md' | 'lg' | 'xl' | 'server';
+
+export const greaterThanOrEqual = (refSize: screenSize) => (size: screenSize) =>
```
brad-decker: Move this into the component, convert to a simple function that takes a prop and compare it to current state size.

src/contexts/ScreenSizeContext.js
```diff
@@ -6,12 +6,27 @@
...
+  screenSizes.indexOf(size) >= screenSizes.indexOf(refSize);
+
+export const lessThanOrEqual = (refSize: screenSize) => (size: screenSize) =>
```
brad-decker: Same as above

src/components/FlexGrid/Row.js
```diff
@@ -1,10 +1,11 @@
 /**
  * @file Row
  * @author Brad Decker <bhdecker84@gmail.com|brad@merlinlabs.com>
+ * @author Ari Frankel <ari.l.frankel@gmail.com|ari@merlinlabs.com>
```
brad-decker: No need to remove all the other js docs just yet (i will tackle in another pr) but lets avoid adding to them (We updated our policies on js docs last week)

src/components/FlexGrid/Column.js
```diff
@@ -1,6 +1,7 @@
 /**
  * @file Column
  * @author Brad Decker <bhdecker84@gmail.com|brad@merlinlabs.com>
+ * @author Ari Frankel <ari.l.frankel@gmail.com|ari@merlinlabs.com>
```
brad-decker: Remove this line per our new js docs policy. The rest of the js docs stuff will be removed in another pr

AriLFrankel:
@brad-decker You are right. This was predicated on what I now see was a misunderstanding of how to use the pattern you introduced with class based components. I am going to close this PR.

PR #215, opened 2018-03-19
feat(Tooltip): Initial implementation

### LINKED ISSUE #47: Make Tooltip component
https://material.io/guidelines/components/tooltips.html

martahj:
Builds out an initial tooltip component

Resolve #47 

Now link: https://docs-kjhyzertdn.now.sh

### DIFF
--- a/src/components/Tooltip.js
+++ b/src/components/Tooltip.js
@@ -1,0 +1,214 @@
+import React from 'react';
+import styled from 'styled-components';
+import { Portal } from './Portal';
+import { Icon } from '../icons/icons';
+
+const MOBILE_MARGIN = 24;
+const DESKTOP_MARGIN = 14;
+const MOBILE_FONT_SIZE = 14;
+const DESKTOP_FONT_SIZE = 10;
+const MOBILE_SIDE_PADDING = 16;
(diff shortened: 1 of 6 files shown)


martahj:
A note on this PR: In the past, we've used [react-tooltip](https://github.com/wwayne/react-tooltip) for our tooltip needs. However, this implementation moves away from that because it is no longer being actively maintained and does not work cleanly with styled-components (see [this](https://github.com/wwayne/react-tooltip/issues/110) issue)

src/components/Tooltip.js
```diff
@@ -0,0 +1,214 @@
+import React from 'react';
+import styled from 'styled-components';
+import { Portal } from './Portal';
+import { Icon } from '../icons/icons';
+
+const MOBILE_MARGIN = 24;
```
brad-decker: Probably not now but sometime soon we need to have a group session where we identify common variables used throughout the material components implementation as well as our own library and move them into the theme or create some variables that are usable globally.

martahj: Good idea. I opened an [issue](https://github.com/MerlinLabs/styled-material-components/issues/226) for this.

PR #216, opened 2018-03-19
fix(Tabs): Fix styled

BennettStaley:
Fixes styled exports on the tab component so it exports the styled
versions of everything.

### DIFF
--- a/src/components/Tabs/Tab.js
+++ b/src/components/Tabs/Tab.js
@@ -39,16 +39,12 @@ const TabComponent = ({ children, className, Icon, index, onClick }) => (
   </RippleTab>
 );
 
-export const Tab = styled(TabComponent)`
+export const Tab = styled(TabComponent) `
   opacity: ${props => (props.selected ? 1 : 0.7)};
   height: ${props => (props.Icon && props.children ? 72 : 48)}px;
 `;
 
-const TabContentContainer = styled.div`
+export const TabContent = styled.div`
   padding: 24px;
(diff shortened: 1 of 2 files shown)


BennettStaley:
I didn't deploy this as its no style change, just some export changes.

src/components/Tabs/Tab.js
```diff
@@ -39,16 +39,16 @@ const TabComponent = ({ children, className, Icon, index, onClick }) => (
...
 );
 
-export const Tab = styled(TabComponent)`
+export const Tab = styled(TabComponent) `
   opacity: ${props => (props.selected ? 1 : 0.7)};
   height: ${props => (props.Icon && props.children ? 72 : 48)}px;
 `;
 
-const TabContentContainer = styled.div`
+const TabContentComponent = ({ children, className }) => (
```
hsarb: this revets the new API styling we were trying to accomplish

BennettStaley: Unless I'm mistaken, its impossible to style that element reliably without calling styled on it. As the component that was exported wasn't a SC, but its child was.

hsarb: So would it make sense then just to export the styled div then?

```
export const TabContent = styled.div`
  padding: 24px;
  width: 100%;
`;
```

BennettStaley: Actually... yes? I think so.

PR #217, opened 2018-03-19
fix(Dialog): Control onClose

dekkofilms:
This allows us to pass in the onClose handler from the parent.

https://docs-qjyesojtti.now.sh

### DIFF
--- a/src/components/Dialog/Dialog.js
+++ b/src/components/Dialog/Dialog.js
@@ -1,4 +1,4 @@
-import React, { Component } from 'react';
+import React from 'react';
 import styled from 'styled-components';
 import elevation from '../../mixins/elevation';
 import { Portal } from '../Portal';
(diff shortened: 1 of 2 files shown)


src/components/Dialog/Dialog.js
```diff
@@ -17,12 +17,12 @@ class DialogComponent extends Component {
...
     if (prevOpen !== nextOpen) return this.setState({ open: nextOpen });
-    if (nextOpen && !this.state.open) this.openModal();
   }
 
-  closeModal = () => this.setState({ open: false });
-
-  openModal = () => this.setState({ open: true });
+  closeModal = () => {
+    this.props.onClose && this.props.onClose();
+    this.setState({ open: false });
```
hsarb: should we execute state changes for handling a callback? Do we want state to resolve before a callback resolves, or should state be prevented from executing if the callback results in a failure?

dekkofilms: I would like to rip out state completely ... I'm having a hard time thinking of a use case where a dialog box is ever uncontrolled.

PR #219, opened 2018-03-19
enhancement(Transitions): Mixin for duration and easing

hsarb:
Mixin to add easing and duration css properties.

Did not create a new build, as it is not currently being used.

Could use `platform` to detect current platform, but would need a list of platform os's that need to be supported.

### DIFF
--- a/src/mixins/transition.js
+++ b/src/mixins/transition.js
@@ -1,0 +1,39 @@
+import platform from 'platform';
+import { css } from 'styled-components';
+
+const isMobile = platform.os.family.match(/^(Android|iOS|Windows Phone)$/);
+
+export const durations = {
+  complex: css`
+    transition-duration: ${isMobile ? 375 : 200}ms;
+  `,
+  standard: css`
+    transition-duration: ${isMobile ? 300 : 150}ms;
+  `,
+  enteringScreen: css`
+    transition-duration: 225ms;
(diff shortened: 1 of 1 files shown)


src/mixins/transition.js
```diff
@@ -0,0 +1,39 @@
+import { css } from 'styled-components';
+
+const isMobile =
+  typeof window.orientation !== 'undefined' || navigator.userAgent.indexOf('IEMobile') !== -1;
```
martahj: [Q] Will this be true for tablets as well?

hsarb: Should be yes.

brad-decker: Can you walk me through this logic and how safe this is to use to determine if the device is mobile?

hsarb: Sure. It's basically just looking to see if in the current browser window.oriententation exists. If so, the browser is most likely mobile.  I'm not sold on the safety of this as I was just using it for now as a quick win. I would prefer to use platform or navigator.userAgent to do the check, but would need a list of supported environments. 

Looks like my concerns are valid as it has been deprecated. 

https://developer.mozilla.org/en-US/docs/Web/API/Window/orientation

Going to update the PR with userAgent checks. Will just check for Android iOS and windows phone for now.

PR #221, opened 2018-03-19
feat(GridList): Implement Sticky Subheaders 

ccm36:
NOTE: subheader background-color styling currently defaults to 'transparent'; may cause some issues on darker backdrops

LINK: https://docs-eubqqiykfd.now.sh/gridlists/


src/components/GridList/GridSubheader.js
```diff
@@ -0,0 +1,21 @@
...
+  </div>
+);
+
+const GridSubheader = styled(GridSubheaderComponent)`
+  text-align: left;
+  top: 0px;
+  height: 48px;
+  width: 100%;
+  padding-left: 16px;
+  position: sticky;
```
martahj: We haven't been checking for browser compatibility so I don't know that we should start now, but position: sticky [isn't supported in IE at all](https://developer.mozilla.org/en-US/docs/Web/CSS/position). It's probably part of a larger question about what browsers we want to support (opened a new issue [here](https://github.com/MerlinLabs/styled-material-components/issues/224)) so I don't know that it should hold up this PR, but I'd be interested to hear others' thoughts on this

ccm36: Thank you for pointing that out, I know Chrome was most recent to adopt `position: sticky;` 
I tested with Chrome, Firefox, and Safari but it escaped me to check IE.  

I was looking into adding logic to displace the previous subheader when a new subheader reaches the top of the window, and ended on `position: sticky;` so as not to overcomplicate.

I'd be happy to refactor should that be the desired route

brad-decker: Thought i was gonna have something awesome to help us out here https://developer.mozilla.org/en-US/docs/Web/CSS/@supports <- this is awesome

Unfortunately it is ALSO not supported by IE :P

ccm36: Nonetheless, still a very cool tool to learn about! Let me know the direction you think would be best to move forward, thanks for the insight

ceciliamvrie:
@petegivens Ive tried looking for this as well, and there seems to be no examples describing what the sticky subheader should look like except for what's on this page https://material.io/guidelines/components/subheaders.html#. 

It also appears from this pr opened on Material UI that the default behavior of a subheader should be sticky. The issue with browser compatibility is also mentioned here. https://github.com/mui-org/material-ui/pull/8194

PR #223, opened 2018-03-19
fix(Table): Fix SortBy

danielstclair:
Tuen sortBy into an arrow function in order to keep "this" context when
checking for props


BennettStaley:
Noted issue: While this functions perfectly normally inside SMC context, once SMC is installed as a module it fails to bind context. Reminiscent of the react import issue we had before.

PR #225, opened 2018-03-20
Refactor(DropdownMenu)

ceciliamvrie:
Refactor DropdownMenu component
to handle callbacks on item selection

Link => https://docs-bruvuxsidn.now.sh

### DIFF
--- a/src/components/DropdownMenu.js
+++ b/src/components/DropdownMenu.js
@@ -33,9 +33,6 @@ const Symbol = styled(ArrowDropDownIcon)`
   fill: #726969;
 `;
 
-/*
- * TODO implement real onchange handler for select field
- */
 export default class DropdownMenu extends Component {
   state = {
     isOpen: false,
@@ -43,7 +40,7 @@ export default class DropdownMenu extends Component {
     selected: this.props.defaultOption || 'select one',
     isChrome: undefined,
   }
-
+  
   /* eslint-disable react/no-did-mount-set-state */
   /*
    * Because Next.js executes its code server-side first,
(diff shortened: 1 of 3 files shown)


src/components/DropdownMenu.js
```diff
@@ -85,9 +83,10 @@ export default class DropdownMenu extends Component {
     const { options, isOpen, selected, isChrome } = this.state;
     return (
       <div onClick={this.toggleSelect} onFocus={this.handleOpen} tabIndex="0">
-        {isChrome === false && (
+        {!isChrome && (
           <Dropdown
             defaultValue={selected}
+            onClick={({ target: { value } }) => this.props.callback(value)}
```
danielstclair: [q] Wouldn't this fire when clicking to open the dropdown?

hsarb: Dropdown is a select, so should it be `onChange`? Also what if callback is not defined? Can the `callback` prop have a more explicit name?

ceciliamvrie: You're so right, for some reason I thought that the `onChange` event wouldn't fire on the default select Dropdown

ceciliamvrie:
@hsarb Thank you for the input, I renamed it to `itemOnClick`, which is a property passed into the Menu component in MaterialUI

The reference to this property is here => http://www.material-ui.com/#/components/menu.
I would love to hear your thoughts on this 👍

src/components/DropdownMenu.js
```diff
@@ -85,9 +83,10 @@ export default class DropdownMenu extends Component {
     const { options, isOpen, selected, isChrome } = this.state;
     return (
       <div onClick={this.toggleSelect} onFocus={this.handleOpen} tabIndex="0">
-        {isChrome === false && (
+        {!isChrome && (
           <Dropdown
             defaultValue={selected}
+            onChange={({ target: { value } }) => this.onSelectMenuItem(value)}
```
martahj: [Q] Just a question, but why is it onChange if not Chrome and a click handler if it is Chrome?

ceciliamvrie: @martahj I did that because I believe the `onChange` event doesn't work for the Menu component, so I added `onClick` for its items instead

martahj:
@ceciliamvrie Could you post a now link to this?

PR #227, opened 2018-03-20
feat(fullscreendialog): add fullscreen dialog

AriLFrankel:
This adds a full screen dialog component. These are intended to
facilitate complex user interactions on mobile that require the full
screen.

https://docs-cdybiwzkwb.now.sh

### DIFF
--- a/pages/dialog.js
+++ b/pages/dialog.js
@@ -1,54 +1,80 @@
 import React, { Component } from 'react';
 import { Button, Dialog, DialogTitle, DialogBody, DialogFooter, ThemeProvider } from '../src';
 
-class ExampleDialog extends Component {
+const ExampleDialog = ({ title, closeDialog, ...dialogProps }) => (
+  <Dialog {...dialogProps}>
+    {Boolean(title) && (
+      <DialogTitle>
+        {title}
+      </DialogTitle>
+    )}
+    <DialogBody>
+      Dialog body goes here
+    </DialogBody>
+    <DialogFooter>
+      <Button onClick={closeDialog}>
+        Cancel
+      </Button>
+      <Button onClick={closeDialog}>
+        Confirm
+      </Button>
+    </DialogFooter>
+  </Dialog>
+);
+
+export default class dialogsPage extends Component {
   state = {
-    open: false,
+    withTitle: false,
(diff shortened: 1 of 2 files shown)


src/components/FullscreenDialog.js
```diff
@@ -0,0 +1,68 @@
...
+    transform: translateX(107%);
+  }
+
+  &.open {
+    transition: transform 0.225s;
+    transform: none;
+  }
+`;
+
+export const FullscreenDialog = ({
```
martahj: [Q] Out of curiosity, what's the reason for creating a new component instead of adding a `fullscreen` option prop on the Dialog component?

martahj:
@AriLFrankel Could you post the now link?

src/components/FullscreenDialog.js
```diff
@@ -0,0 +1,68 @@
...
+  top: 0;
+  overflow: hidden;
+  position: absolute;
+  transition: transform 0.195s;
+  width: 100%;
+  will-change: transform;
+
+  &.left {
+    left: 0;
+    transform: translateX(-107%);
```
dekkofilms: [Q] I'm just curious about this number 107%? Why that specifically ... just wanted to understand! 👍

AriLFrankel: I borrowed that from the Drawer, and my understanding is that any number above 100% would work to push this off screen in that direction. 👍

src/components/FullscreenDialog.js
AriLFrankel: I wasn't sure at first because it's very different than the other dialogues, and because it should slide in from the side like a drawer rather than appear like a modal, but I guess that it ultimately is a dialog and therefore should land in that world. I can refactor this.

AriLFrankel:
@dekkofilms @martahj this is ready for re-review

dekkofilms:
@AriLFrankel Do you have an updated now link? 😄

dekkofilms:
@AriLFrankel I can't seem to close the dialog by clicking the darkened background ... is that the same for you?

AriLFrankel:
Clearly on it today. Try with the updated link

PR #230, opened 2018-03-21
feat(generateBreakpoint): export generateBreakpoit

AriLFrankel:
Export generateBreakpoint from the defaultTheme file in order to make
it easier to provide custom breakpoints to the theme from user land.


AriLFrankel:
yes

PR #236, opened 2018-03-22
feat(Icons): Add Icons

petegivens:
- Arrow Back
 - Arrow Forward


src/icons/ArrowForwardPath.js
```diff
@@ -0,0 +1,5 @@
+import React from 'react';
+
+export const ArrowBackPath = () => (
```
AriLFrankel: ArrowForward*

petegivens: fixed

PR #239, opened 2018-03-24
refactor(MenuList): Add tab scrolling

ceciliamvrie:
Added the ability to tab through the MenuList
without tabbing out of the Menu

new link =>https://docs-uxodthnqtk.now.sh!

### DIFF
--- a/src/components/Menu/MenuItem.js
+++ b/src/components/Menu/MenuItem.js
@@ -6,15 +6,16 @@ const StyledListItem = styled.li`
 `;
 const RippleListItem = withRipple(StyledListItem);
 
-const MenuItemComponent = ({ className, children, onClick }) => (
+const MenuItemComponent = ({ className, children, onClick, onKeyDown }) => (
   <RippleListItem
     className={`${className} smc-menu-list-item`}
     onClick={onClick}
-    onKeyPress={onClick}
+    onKeyPress={onKeyDown}
     tabIndex="0">
     {children}
   </RippleListItem>
 );
+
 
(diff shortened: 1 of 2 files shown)


martahj:
Looks good!

One potential enhancement might be having it return to the first element if you're focused on the last element and hit 'tab' again - thoughts?

ashwoodall:
Just a [Q] - how would this affect keyboard accessibility? The original intent here was someone with visual or motor impairments could still navigate through the page with the keyboard. Does this allow for some other way to navigate outside of the menu without a mouse? I also noticed that now `shift + tab` doesn't allow the previous item to be selected. 

https://webaim.org/techniques/keyboard/

ceciliamvrie:
@martahj That is a good idea, but in Material UI the focus stays at the last element. Are we trying to mimic Material spec to that such extent?

ceciliamvrie:
@ashwoodall So far, pressing `esc` would allow the user to close the menu if they wish to navigate to other elements. I can provide a check to make sure that the menu is open ignorer for tab navigation to work.

ashwoodall:
As long as there is still a way to progress to the next tabindex element after esc is pressed (doesn't move focus back to the beginning) and we don't lose the `shift + tab` to move to previous element functionality I'm fine with this 👍

ceciliamvrie:
@ashwoodall @martahj Cool! I added `shift + tab` and I made sure that `esc` blurs the current active element so that the user will escape from the Menu without moving their focus back to the beginning.

PR #240, opened 2018-03-26
feat(Snackbar): Add Mobile Snackbar

petegivens:
- utilize withScreenSize context provider to set a prop `mobile` on xs and sm screen sizes
- `mobile` prop used for changing styles conditionally on smaller screen sizes
- add `multiline` prop to indicate that a snackbar should take up two lines instead of one. only works when `mobile` prop has been triggered

Did not add Action Button functionality to snackbar at this time.

### DIFF
--- a/pages/_document.js
+++ b/pages/_document.js
@@ -9,11 +9,12 @@ export default class MyDocument extends Document {
     return (
       <html lang="en">
         <Head>
+          <meta name="viewport" content="width=device-width, initial-scale=1.0" />
           <title>Styled Material Components</title>
           {styleTags}
         </Head>
         <body style={{ overflow: 'visible' }}>
-          <div className='root' style={{ width: '100%' }}>
+          <div className="root" style={{ width: '100%' }}>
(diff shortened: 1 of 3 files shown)


petegivens:
https://docs-hdnxstuuhq.now.sh/snackbar

martahj:
Nice work!

Any ideas why there's a small (~1px) gap to the left of the mobile snackbar?
screen shot 2018-03-26 at 4 36 33 pm

petegivens:
@martahj Thanks, I didn't notice that. I think it had to do with how the snackbar was positioning itself by setting `left: 50%` and using `transform: translateX(-50%)`.  To correct, when `mobile` prop is true I set `left: 0` and remove the `transform`.

petegivens:
I want to add one more change before this gets merged. I added a meta tag in the _document head to make the demo pages behave correctly on a mobile device or when emulating one in chrome dev tools.

petegivens:
addresses #237

PR #241, opened 2018-03-27
Textfield changes

dekkofilms:
This has two commits:

This checks if the input isControlled. Also making sure if the input is controlled that the floating label would work if the value passed in has a string with the length greater than 0.

This adds into the default theme the text field color that we can now control from the theme object.

Now link: https://docs-xtngblqdhp.now.sh/

### DIFF
--- a/src/components/TextField.js
+++ b/src/components/TextField.js
@@ -4,6 +4,12 @@ import React, { PureComponent } from 'react';
 import DropdownMenu from './DropdownMenu';
 
 class TextFieldComponent extends PureComponent {
+  constructor(props) {
+    super(props);
+
+    this.isControlled = props.value != null;
+  }
+
   state = {
     text: this.props.defaultValue || '',
     focus: false,
@@ -22,17 +28,22 @@ class TextFieldComponent extends PureComponent {
   textArea = null;
 
   onChange = (e) => {
-    this.props.onChange && this.props.onChange(e);
-    const text = e.target.value;
-    const isInvalid = this.props.validator && !this.props.validator(text);
-    const isEmptyButRequired = this.props.required ? !e.target.value : false;
-    const newHeight = this.props.textarea ? this.textArea.scrollHeight : '';
+    if (this.props.onChange) {
+      this.props.onChange(e);
+    }
 
-    this.setState({
-      text,
-      error: this.props.error || isInvalid || isEmptyButRequired,
-      height: !this.props.multiline || this.textArea.value === '' ? '100%' : newHeight,
-    });
+    if (!this.isControlled) {
+      const text = e.target.value;
+      const isInvalid = this.props.validator && !this.props.validator(text);
+      const isEmptyButRequired = this.props.required ? !e.target.value : false;
+      const newHeight = this.props.textarea ? this.textArea.scrollHeight : '';
+
+      this.setState({
+        text,
+        error: this.props.error || isInvalid || isEmptyButRequired,
+        height: !this.props.multiline || this.textArea.value === '' ? '100%' : newHeight,
+      });
+    }
   };
 
   onFocus = (e) => {
@@ -52,6 +63,9 @@ class TextFieldComponent extends PureComponent {
     );
 
     const hasError = Boolean(this.state.error || this.props.error || this.props.errorText);
+
+    // Boolean to add to floating label logic
+    const hasValue = (this.isControlled) && this.props.value.length > 0;
     return (
       <div className={`${this.props.className} smc-text-field-container`}>
         <Suffix>{this.props.suffix}</Suffix>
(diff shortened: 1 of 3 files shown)


src/components/TextField.js
```diff
@@ -4,6 +4,12 @@ import React, { PureComponent } from 'react';
 import DropdownMenu from './DropdownMenu';
 
 class TextFieldComponent extends PureComponent {
+  constructor(props) {
+    super(props);
+
+    this.isControlled = props.value != null;
```
martahj: I think I'm concerned about not letting users pass in `null` for the value. It makes sense to check against `undefined` because it's not possible to distinguish between someone passing in the value `undefined` and not someone not passing in a prop, but it might be confusing for a user if they explicitly pass in `null` for a value and lose control of the textfield.

martahj:
@dekkofilms Could you share a now link?

dekkofilms:
@martahj yep! sorry, in description now.

src/components/TextField.js
```diff
@@ -52,6 +63,9 @@ class TextFieldComponent extends PureComponent {
     );
 
     const hasError = Boolean(this.state.error || this.props.error || this.props.errorText);
+
+    // Boolean to add to floating label logic
+    const hasValue = (this.isControlled) && this.props.value.length > 0;
```
martahj: [REQ] Could you rename this to something like `hasControlledValue`?

src/components/TextField.js
dekkofilms: So, would you think only checking for `undefined` would suffice, with a `props.value !== undefined`. This will allow null to come through.

Sorry, I meant to change this when we discussed earlier and must have gotten distracted.

petegivens: I would agree with that. the more common pattern is to omit a prop that's not being used rather than passing it null.

src/theme/defaultTheme.js
```diff
@@ -89,6 +89,9 @@ export const defaultTheme = {
     disabled: black.alpha(0.38).toString(),
     icon: black.alpha(0.38).toString(),
   },
+  inputColors: {
+    textField: '#029FB1',
+  },
```
petegivens: [Q] What's the reason for adding this as a separate theme color? Doesn't material spec dictate that active text fields should be the primary theme color?

dekkofilms: It's up for discussion for sure ... the way we have primary set up in our repo is not going to work.

src/components/TextField.js
dekkofilms: I think we do want to pass it null though.

src/theme/defaultTheme.js
ashwoodall: I'm running into the same issue @dekkofilms is here - when red is the primary color for a company it is VERY confusing when implementing textfields (looks like an error, have to manually override each field to get it to the correct color, etc). 

Thinking about it from a Material Spec standpoint though, wouldn't the textfield color be our primary color? Maybe we should change the way we're implementing this internally. Is there anywhere where we actually want our primary color to be red with SMCs?

dekkofilms: @ashwoodall I know we are also using a lot of primary buttons, which we do want to be that red color. womp. womp. I guess it might have to be a bigger change in our pages. I don't mind reverting this at all.

petegivens: Can this be solved by just tweaking the theme we're using in CMS?

src/components/TextField.js
petegivens: @dekkofilms you're saying the user might pass in `value={null}` to signify that it won't be controlled?

src/theme/defaultTheme.js
ashwoodall: I'm wondering this too @petegivens. I'm good with whichever seems best! It would be super helpful to have a better way than what's going on currently though

petegivens: @ashwoodall I feel pretty strongly that keeping our colors in the theme and rendering buttons based on `primary` and `accent` colors is the way to go. But TBH I haven't run into your issues, so we may need an alternative approach, e.g. adding a `buttonColor` to the theme in our ThemeProvider in CMS.

src/components/TextField.js
dekkofilms: I think @martahj is suggesting that potential use case, but we want it to still be potentially controlled if it does have a null value due to how some fields resolve in graphql.

Right now this approach will not make it controlled if they do pass null. What this is helping is if the user passes in an initial value of `''`, which is a lot of cases ... we still want it to be known as a controlled component.

martahj: I would think that if the user passes *any* value, including `''` or `null`, they are probably trying to control the component, and if they don't pass a value (so `props.value` is undefined) they are probably not trying to control the component... thoughts?

src/theme/defaultTheme.js
dekkofilms: I really like that ... upcoming commit will do that, and then open to more opinions and thoughts!

dekkofilms:
@martahj @petegivens @ashwoodall issues addressed!

src/theme/defaultTheme.js
petegivens: I don't have a problem with that approach, but I think the best way to deal with it is to wrap the component or page in a `<ThemeProvider>` that overrides the primary color.

PR #242, opened 2018-03-27
feat(Icon): Add All Material Icons

petegivens:
- refactored Icon component
- added a script that recursively finds all relevant svg files from
  material-design-icons repo and generates a react component based on
  each file it finds
- executed script and generated icon components
- no breaking changes to existing Icon component, but does not extend the
  convention of receiving a named icon prop

- New Icon Props:
 -- fill: fill color
 -- size: number of pixels in the icon's height and width

----------
This adds all icons in their named standalone instances, e.g. `<ArrowBackIcon />`. It does not extend the `<Icon icon="icon_name" />` pattern. My initial attempts to do that were causing a huge client package whenever `Icon` was imported. I'll keep trying to figure out how to do that using conditional imports.


petegivens:
The demo page needs some work:
 https://docs-pqygvfcvfl.now.sh/icons

scripts/generate-icons.js
```diff
@@ -0,0 +1,185 @@
...
+recursive(INPUT_DIR, options, (err, files) => {
+  if (err) throw err;
+  files.forEach((file) => {
+    fs.readFile(file, 'utf-8', (err, data) => {
+      if (err) throw err;
+      const paths = getSvgPaths(data);
+      if (paths && file) {
+        writeComponent(paths, file);
+        addToIndex(file);
+        // addToMap(file);
```
martahj: [Q] Is the intention to eventually add this function back in?

petegivens: yep. I think the `<Icon icon="icon_name" />` api is really nice, but i wasn't able to make it work without making `Icon` a 30MB import.

martahj: lol fair 😛

petegivens: @martahj that's also why i'm exporting the path component as default, because it would be necessary for them to be default exports if we use Next.js dynamic.

PR #257, opened 2018-03-30
refactor(Contexts): Use official React 16.3 api

brad-decker:
This removes the polyfill in favor of the real thing. There is a
breaking change in that I have reverted the 'functional hoc' style for
the withScreenSize and withRowState contexts because it comes with side
effects that are unintentional with other apis that were exposed by
react 16.3 update. Namely this means the forwardRef api.

BREAKING CHANGE:
Usage of withScreenSize/withRowState must be updated.

### DIFF
--- a/src/contexts/ScreenSizeContext.js
+++ b/src/contexts/ScreenSizeContext.js
@@ -6,23 +6,12 @@
  * access the current screenSize. This is implmented using respondable.
  * All breakpoints are adjustable by the end user.
  */
-import React, { Component } from 'react';
-import createReactContext from 'create-react-context';
+import React, { Component, createContext, forwardRef } from 'react';
 import { withTheme } from 'styled-components';
 import respondable from 'respondable';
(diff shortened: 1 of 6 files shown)


src/contexts/ScreenSizeContext.js
```diff
@@ -140,8 +129,15 @@ class ScreenSizeContextBase extends Component {
...
+export function withScreenSize(WrappedComponent) {
+  const ScreenSizeAwareComponent = (props, ref) => (
+    <ScreenSizeConsumer>
+      {screenSizeState =>
+        <WrappedComponent {...props} screenSizeState={screenSizeState} ref={ref} />
+      }
+    </ScreenSizeConsumer>
+  );
+  const name = WrappedComponent.displayName || WrappedComponent.name;
+  ScreenSizeAwareComponent.displayName = `screenSizeAware(${name})`;
```
petegivens: this is tight

PR #258, opened 2018-04-04
feat(Avatar): Add initials

ashwoodall:
This PR adds the ability for a name prop to be passed to Avatar and creates initials if an image src is not defined. It handles passed first + last name, first (single) name, initials (2 chars), or no name prop.

### DIFF
--- a/src/components/Avatar.js
+++ b/src/components/Avatar.js
@@ -2,8 +2,22 @@ import React from 'react';
 import { Circular } from './Circular';
 import { Image } from './Image';
 
+const formatProps = ({ name, children }) => {
+  if (!name) {
+    return children;
+  } else if (name.length <= 2) {
+    return name;
+  }
+
+  const nameArr = name.split(' ');
+  const initials = nameArr.length === 1 ? nameArr[0][0] : nameArr[0][0] + nameArr[1][0];
+
+  return initials.toUpperCase();
+};
+
 export const Avatar = Circular.extend.attrs({
-  children: props => (props.src ? <Image src={props.src} /> : props.children),
+  children: props =>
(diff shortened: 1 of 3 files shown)


src/components/Avatar.js
```diff
@@ -2,8 +2,22 @@ import React from 'react';
...
+  const nameArr = name.split(' ');
+  const initials = nameArr.length === 1 ? nameArr[0][0] : nameArr[0][0] + nameArr[1][0];
+
+  return initials.toUpperCase();
+};
+
 export const Avatar = Circular.extend.attrs({
-  children: props => (props.src ? <Image src={props.src} /> : props.children),
+  children: props =>
+    (props.src ? <Image src={props.src} /> : createInitials(props.name, props.children)),
```
martahj: This is kind of a NIT and you don't really have to do this, but I find it somewhat confusing to have the logic for returning the children combined with the logic for creating the initials. The function of `createInitials` might be clearer if it was broken into two functions, one that created the initials and one that determined whether to return initials or children... or I guess if `createInitials` was renamed

martahj:
This is looking good!

Since we're temporarily unable to use Now, could you please do the following:
* Pull in the latest change
* Add a working example of this being used to a demo page
* Run `yarn dev`
* Verbally verify that this is behaving as expected

src/components/Avatar.js
ashwoodall: For the sake of keeping the ternary clean I destructured props and renamed the function to `formatProps`. LMK if this makes more sense, and if not I don't mind breaking it out into separate functions.

martahj:
Such a beautiful demo page. Look at that gradient behind 'hi'.

PR #259, opened 2018-04-04
Refactor(Tooltip): Restyle Tooltip

ceciliamvrie:
Made some small style edits to the tooltip component.

### DIFF
--- a/src/components/Tooltip.js
+++ b/src/components/Tooltip.js
@@ -3,8 +3,8 @@ import styled from 'styled-components';
 import { Portal } from './Portal';
 import { Icon } from '../icons';
 
-const MOBILE_MARGIN = 24;
-const DESKTOP_MARGIN = 14;
+const MOBILE_MARGIN = 16;
+const DESKTOP_MARGIN = 7;
 const MOBILE_FONT_SIZE = 14;
 const DESKTOP_FONT_SIZE = 10;
 const MOBILE_SIDE_PADDING = 16;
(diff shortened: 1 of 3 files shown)


ceciliamvrie:
@martahj Should I also add in @nataliaisabelle for QA?

src/components/Tooltip.js
```diff
@@ -18,7 +18,8 @@ const ROBOTO_RATIO = 0.51; // This is a guess
 export const TooltipPortal = styled(Portal)`
   && {
     box-sizing: border-box;
-    background: rgba(0, 0, 0, .9);
+    background: #616161;
```
martahj: [Q] What was the reason for specifying a solid color rather than an opacity? The spec mentions that tooltips should have an opacity of 90% https://material.io/guidelines/components/tooltips.html#tooltips-interaction

ceciliamvrie:
@martahj We're going to need to talk to @brad-decker about upgrading

src/components/Tooltip.js
ceciliamvrie: I happened to glance over that in the spec, but I just added an `opacity` property. Would that be good, or would it be better to convert this back to `rgba`?

martahj: The problem with `opacity` is that it impacts all of the component's children as well, which can look weird (imo) for fine detail such as text or images. `rgba` only impacts the element that it is attached to. I'm 👍 with using opacity if @nataliaisabelle prefers it because it can always be overridden if the use case requires, however.

https://stackoverflow.com/questions/13385492/css-opacity-vs-rgba-which-one-is-better

ceciliamvrie:
@martaj sounds good!

danieltisdale:
@ceciliamvrie Two things to add that were discussed with @nataliaisabelle:

- [ ] Hover on desktop needs to show default cursor

- [ ] Make the tooltip tabbable

ceciliamvrie:
@danieltisdale and @nataliaisabelle this pr is ready for review!

PR #260, opened 2018-04-04
fix(Column): fix breaking changes from ScreenSizeContext

petegivens:
This PR fixes a bug that was breaking most of the SMC pages where Column was not able to be converted to a styled component due to some effect of returning a forwardRef instead of the actual component. Per @brad-decker, the forwardRef was only a failsafe so we're removing it for now.

### DIFF
--- a/src/components/FlexGrid/Column.js
+++ b/src/components/FlexGrid/Column.js
@@ -26,9 +26,8 @@ import { withRowState } from './Row';
  * @param {String} key             The subkey to return from the prop object
  */
 const getResponsiveProp = (props, layout, screenSize, lowerScreenSize, key) => {
-  const responsiveProp = props[screenSize]
-    || props[lowerScreenSize]
-    || props[layout.defaultScreenSize];
+  const responsiveProp =
+    props[screenSize] || props[lowerScreenSize] || props[layout.defaultScreenSize];
   if (!responsiveProp) return null;
   if (key) return responsiveProp[key];
   return responsiveProp;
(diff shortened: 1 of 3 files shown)


martahj:
Cool!

In terms of the forwardRef, I've opened a PR that would add styled-components support for these https://github.com/styled-components/styled-components/pull/1658. It was working via symlink but who knows what they'll think of it. It doesn't sound like the forwardRef is super important to the Column though so it sounds like we should move forward with your PR regardless, as this is a blocker for all SMC work.

I'd ask you to post a Now link but that's not possible right now.. can you just verbally verify that it's working locally? 😉

petegivens:
@martahj lol yeah, I tried to build one, but...

I have verified that all the SMC pages are working locally.

brad-decker:
Who's this guy that keeps breaking things??? <_< 

Good work @petegivens cleaning this up, and woohoo @martahj hope that gets merged!

petegivens:
Actually, hold off on merging for a minute please. Just noticed a bug with the snackbar.

petegivens:
@brad-decker @martahj There's another issue cropping up. All I've determined for sure so far is that Snackbar breaks if we pass a `ref` into it from ScreenSizeContext. Flex-Grid breaks if ScreenSizeContext _doesn't_ pass it a `ref`. I've tried subbing `innerRef` instead but it made no difference. The error is:
```
TypeError: Cannot add property current, object is not extensible
    at safelyDetachRef (http://localhost:8081/_next/1522889688587/commons.js:9650:21)
    ...etc
```
       ¯\_(ツ)_/¯

petegivens:
I was getting the above-mentioned error from the dev server and 
`Warning: Unexpected ref object provided for Styled(SnackbarComponent). Use either a ref-setter function or React.createRef().` in the console. 

I resolved by removing the `ref` that ScreenSizeContext was passing down, but I'm not sure why it wasn't being recognized as a valid ref. 

Additionally, withRowState is throwing this warning: `Stateless function components cannot be given refs. Attempts to access this ref will fail.` 

Just going to remove the `ref`s from both for now, what do you think?

martahj:
@petegivens Removing the refs sounds fine to me

PR #262, opened 2018-04-05
fix(Dialog): No scroll on body when open

dekkofilms:
This prevents the body from scrolling when the dialogs are open.

This is not up to date quite yet with the latest changes, I didn't pull down till those fixes were resolved. Will rebase and change the `componentWillReceiveProps` method as needed.


src/components/Portal/Portal.js
```diff
@@ -41,6 +57,7 @@ export class Portal extends Component {
   }
 
   render() {
+    console.log('hit render');
```
petegivens: [Q] Leftover?

dekkofilms: lol yea ... i just pushed up that change when you made the comment. I have to unfortunately close and open this PR again because of CI. Tag you on that one!

PR #263, opened 2018-04-05
feat(icon): add onClick

petegivens:
This PR allows Icon to accept an `onClick` function. Previously, we've been wrapping the icons in divs to accommodate onClick functions.

### DIFF
--- a/pages/icons/index.js
+++ b/pages/icons/index.js
@@ -11,7 +11,6 @@ import {
   ListItemText,
   ListSubheader,
   ThemeProvider,
-  withRipple,
 } from '../../src';
 import { iconList } from './assets';
 import { AddIcon, FileDownloadIcon, FileUploadIcon } from '../../src/icons';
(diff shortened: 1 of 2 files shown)


martahj:
Sweet!

Since we're temporarily unable to use Now, could you please do the following:
* Pull in the latest changes
* Run `yarn dev`
* Verbally verify that this is behaving as expected

petegivens:
@martahj yes, I have pulled down latest changes and verified that it's working. One thought--while I'm add it, do you think there's value in also adding `onMouseUp` and `onMouseDown` props?

martahj:
@petegivens What would be the use case?

petegivens:
@martahj i don't have any specific use-case in mind, just thought it might give more granular control to the user. but then i guess we could talk about onKeyPress and all kinds of stuff so maybe there's no need :)

martahj:
I don't see a huge need for it, though it probably wouldn't hurt anything

petegivens:
Let's just leave it as-is until we find a need for it

PR #264, opened 2018-04-05
fix(Dialog): No scroll on body when open

dekkofilms:
This prevents the body from scrolling when the dialogs are open.

### DIFF
--- a/pages/dialog.js
+++ b/pages/dialog.js
@@ -41,7 +41,8 @@ export default class dialogsPage extends Component {
   render() {
     return (
       <ThemeProvider>
-        <div>
+        <div style={{ height: '200vh' }}>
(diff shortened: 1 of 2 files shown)


martahj:
@dekkofilms Since we don't have Now (for now), @nataliaisabelle will need to QA locally

pages/dialog.js
```diff
@@ -41,15 +41,16 @@ export default class dialogsPage extends Component {
   render() {
     return (
       <ThemeProvider>
-        <div>
+        <div style={{ height: '200vh' }}>
```
ashwoodall: NIT: why not 100vh?

dekkofilms: I just wanted a lot of room to scroll with on the page ... no strong reason for the specific number. I'm sure once we have real docs built out with examples and info, the need to change to the height for testing wouldn't be needed.

PR #265, opened 2018-04-05
feat(menu): open to space

AriLFrankel:
Open menu to where there is space for it. If there is not room for it below open vertically. If there is not room to the right open to the left.

### DIFF
--- a/src/components/Menu/Menu.js
+++ b/src/components/Menu/Menu.js
@@ -1,5 +1,6 @@
 import React, { Component } from 'react';
-import styled, { css } from 'styled-components';
+import styled from 'styled-components';
+import debounce from 'lodash.debounce';
 import elevation from '../../mixins/elevation';
 import MenuList from './MenuList';
 import MenuItem from './MenuItem';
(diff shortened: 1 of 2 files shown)


src/components/Menu/Menu.js
```diff
@@ -1,65 +1,86 @@
...
+    const overBottom = top + menuRect.height > window.innerHeight;
+    const overRight = left + menuRect.width > window.innerWidth;
+    this.menu.style.top = `${anchorRect.top -
+      (overBottom ? menuRect.height - anchorRect.height : 0)}px`;
+    this.menu.style.left = `${anchorRect.left -
+      (overRight ? menuRect.width - anchorRect.width : 0)}px`;
+  }, 0);
 
   render() {
+    this.recalculatePosition();
```
martahj: [Q] Why is this in render? It's not changing state so it's not triggering a rerender, but still feels a little less intuitive than calculating it on scroll

AriLFrankel: @martahj If we only trigger `recalculatePosition` on scroll or resize events then the menu will not be placed correctly when it is first opened before any of those events take place

AriLFrankel: I like that a lot!

martahj:
Looks pretty good to me!

In lieu of a Now app, please verbally verify that this is working locally 😅

AriLFrankel:
I hereby verify that it works on my machine lol

PR #274, opened 2018-05-01
fix(TextField): Implemented tabIndex

ceciliamvrie:
Provided an optional tabIndex prop so the TextFields in forms
can be tabbed through in a certain order


ceciliamvrie:
Also requesting @danielstclair's opinion on this just incase if I missed any pitfalls

martahj:
@ceciliamvrie Could you get @danieltisdale or @nataliaisabelle to QA this locally (either on your machine or theirs) as well?

danielstclair:
I think this looks good. Any concerns here would be in the implementation side of things once we're using this tabIndex prop.

martahj:
@danieltisdale has verbally approved this

PR #280, opened 2018-05-15
Improve Build/CI

martahj:
Sets up flow and jest, and updates the eslint config.


.circleci/config.yml
```diff
@@ -16,5 +16,6 @@ jobs:
           paths:
             - "/node_modules"
       - run: yarn run lint
+      - run: yarn run flow
+      - run: yarn run test
```
brad-decker: [REQ] We need to move danger ahead of hte other two steps so we can still get deployments in the interim until all green.

martahj: Good call

PR #281, opened 2018-05-30
update dependencies

BennettStaley:
This is my initial pass at updating SMC to the new version of next, and styled-components.

MAJOR WIP. Lots of copy and paste, needs to trim dependencies.


src/components/Menu/Menu.js
```diff
@@ -49,7 +49,7 @@ class MenuComponent extends Component {
         className={`${className} smc-Menu`}
         ref={(ref) => {
           this.menu = ref;
-          this.recalculatePosition();
+          debounce(this.recalculatePosition, 0);
```
petegivens: is this just a style choice or are there any implications of calling it here rather than baking it into the method?

BennettStaley: it was throwing a now is not defined error on hot reload. This was the remedy for that. That's all I know.

brad-decker:
@BennettStaley have you tried using link to use this locally? I think we can debug it a little bit ot make sure it doesn't break anything.

petegivens:
YALC IT: https://www.npmjs.com/package/yalc

BennettStaley:
@brad-decker 

The time has come. You know what must be done.

PR #285, opened 2018-05-31
feature(switch): switch improvements

petegivens:
* refactor Switch
* add onClick functionality
* add `radialReaction` prop to disable ripple effect
* add utility mixin for deriving light opaque colors from a base
* add `color` argument to `ripple` mixin


src/components/Switch.js
```diff
@@ -0,0 +1,109 @@
...
+import elevation from '../mixins/elevation';
+import ripple from '../mixins/ripple';
+import { deriveLightColor } from '../mixins/deriveLightColor';
+
+class SwitchComponent extends PureComponent {
+  state = {
+    on: false,
+  };
+
+  componentDidMount = () => {
```
AriLFrankel: What's the reason for setting state from props here as opposed to when you initialize state?

src/components/Switch.js
```diff
@@ -0,0 +1,109 @@
...
+    this.setState({
+      on: this.props.on,
+    });
+  };
+
+  componentDidUpdate = (prevProps, prevState) => {
```
AriLFrankel: My understanding is that your intention is to read `on` from props exclusively when it's available and not from state. If that's correct: not necessarily a req, but i think you could get away without this if you just check for props.on inside handleOnClick, right?

petegivens: Right. I didn't want to put it in the handleOnClick because I want to allow for the component to be controlled from outside. (e.g.--a user is filling out a form, and selecting one option requires that another one be selected). But we can check for that in the render call and then eliminate both CDM and CDU 👍

src/components/Switch.js
```diff
@@ -0,0 +1,99 @@
...
+      this.setState(prevState => ({
+        on: !prevState.on,
+      }));
+      this.props.onChange && this.props.onChange();
+    }
+  };
+
+  render() {
+    // determine if controlled or un-controlled
+    const on = this.props.on ? this.props.on : this.state.on;
```
AriLFrankel: i think you'd want to check against `this.props.on === undefined` here in case false is passed in

src/components/Switch.js
```diff
@@ -0,0 +1,99 @@
...
+import ripple from '../mixins/ripple';
+import { deriveLightColor } from '../mixins/deriveLightColor';
+
+class SwitchComponent extends PureComponent {
+  state = {
+    on: false,
+  };
+
+  handleOnClick = () => {
+    if (!this.props.disabled) {
```
AriLFrankel: Not necessarily a req, might be better to check for props.on at this stage so that you could skip the setState call and the resulting rerender, as opposed to checking inside render.

src/components/Switch.js
petegivens: 🤦‍♂️

src/components/Switch.js
petegivens: good call

petegivens:
@AriLFrankel ready freddy

src/components/Switch.js
```diff
@@ -0,0 +1,111 @@
...
+  state = {
+    on: false,
+  };
+
+  /**
+   * TODO: add an invisible input element to switch.
+   * This will improve behavior of the switch if being used uncontrolled.
+   * It will also be good for a11y.
+   */
+  isControlled = () => this.props.on !== undefined;
```
AriLFrankel: Why does this need to be a function?

petegivens: Because I want to re-check on every update, not just when the component is initialized. I could keep it in state and add a CDU, but this seemed simpler. 

I honestly don't know if there are any valid use cases for this, but I wanted to make it more flexible just in case. One (possibly contrived) example: a user is letting the checkbox be uncontrolled and just using its onChange function. However, given a certain state, the user might force the checkbox to be `disabled` and `on`.

AriLFrankel: SGTM

PR #291, opened 2018-06-06
fix(Menu): find offsetParent

AriLFrankel:
Account for offsetParent of the anchorEl when computing the menu's
position.

Protect against accidental openings by checking that a clicked element
is an ancestor of the anchor element and not just the anchor element
itself.

### DIFF
--- a/src/helpers.js
+++ b/src/helpers.js
@@ -1,0 +1,12 @@
+
+// This is a helper for determining if there is an ancestral relationship between two components
+export const isDescendant = (parent, child) => {
+  let node = child.parentNode;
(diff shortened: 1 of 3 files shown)


src/helpers.js
```diff
@@ -0,0 +1,12 @@
+
+// This is a helper for determining if there is an ancestral relationship between two components
+export const isDescendant = (parent, child) => {
+  let node = child.parentNode;
+  while (node != null) {
```
martahj: The rare double equals

AriLFrankel: true, this could be !==

PR #295, opened 2018-06-08
feat(Menu): attachment styles

AriLFrankel:
Attachment styles: everyone's a little different.

Accept props for bottom and right attachments, as well as upwards and downwards opening

attachBottom: align top of menu with bottom of anchor
openUp: open the menu upwards, align the bottom of the menu with the bottom of the anchor
openLeft: open the menu leftwards, align the right of the menu with the right of the anchor

Defaults: attaches Top Left, opens Right Down

### DIFF
--- a/pages/menus.js
+++ b/pages/menus.js
@@ -1,8 +1,29 @@
 import React, { Component } from 'react';
-import { ThemeProvider, Button, MenuItem, Menu } from '../src';
+import { ThemeProvider, Button, MenuItem, Menu, Switch } from '../src';
+import styled from 'styled-components';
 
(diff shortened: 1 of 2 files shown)


AriLFrankel:
@martahj added examples

martahj:
Very cool. I think I'm having trouble seeing how these differ because they're all at the bottom of the page, could we walk through these tomorrow?

martahj:
When the menu doesn't cover the button, clicking on the button closes and then re-opens the menu. Is this the expected behavior?

AriLFrankel:
Good catch. That was a regression. Fix pushed

PR #301, opened 2018-06-21
feat(chip): update focus style

petegivens:
This PR is a small iteration towards matching MD2 spec for Chip component.

This adds the correct styles for hover, focus, and clicked states. 

After discussion with @nataliaisabelle on our interpretations of the MD2 spec, these are the implemented changes:

- Chips that are clickable have a hover state, focused state, and clicked state
  - MD2 clicked state is what was previously the MD1 focused state
- Chips that are removable but not clickable have a focused state but no hover state
  - The Remove Icon has a hover state
- Chips that are neither clickable nor removable have none of the above

### DIFF
--- a/src/components/Chip.js
+++ b/src/components/Chip.js
@@ -1,31 +1,61 @@
 import React, { PureComponent } from 'react';
-import styled from 'styled-components';
+import styled, { css } from 'styled-components';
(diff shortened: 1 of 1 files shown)


martahj:
@nataliaisabelle Could you QA this PR?

PR #303, opened 2018-06-29
fix(Menu): suppress flashing

AriLFrankel:
This prevents the Menu component from displaying until it is in the
correct location. This fixes a behavior where the menu flashes on
screen before moving to the desired position


martahj:
@nataliaisabelle Could you QA this PR?

PR #314, opened 2018-08-02
Fix/Regressions

ceciliamvrie:
### Summary of changes:
 - Addressed some of the bugs mentioned in (INSERT ISSUE)
 - Added proper flow typing along with minor refactors to the component pages

NOTE: The components I modified should all be listed below

### Behaviors that should be QA'd:
 - There are no regressions for any of the components I modified.

Some of the components modified are used for building other components in smc, 
such as the portal. I will try to list some components that happen to be indirectly effected by my modifications as well to insure that there are no further regressions that will take place. 

NOTE: If the current un-updated version of the smc website is still up, that would be a
good place to measure wether I have fixed the bugs or not

### Components Modified: 
- BottomSheets: Modal should toggle off on clicking outside of it
- Checkboxes: Should be able to pass custom styles / checkmarks in checkboxes
- Drawers: Temporary drawers should toggle off on clicking outside

- Dialogs: Added basic flow typing
- Sliders: Added flow typing
- TextFields: Added flow typing
- Snackbars: Added flow typing

### Components Modified that Impact Others: 
- Portal: Is used by bottom sheets, drawers, and dialogs

### Github Issues closed by this PR:
 - Resolves # (INSERT ISSUE)

### DIFF
--- a/pages/bottomsheets.js
+++ b/pages/bottomsheets.js
@@ -24,19 +24,17 @@ export default class PortalPage extends Component<
     persistent: false
   };
 
-  toggleModal = (): void =>
-    this.setState(
-      (prevState: PortalPageStateType): PortalPageStateType => ({
-        modal: !prevState.modal
-      })
-    );
+  toggleModal = (): void => {
+    this.setState(({ modal }: PortalPageStateType) => ({
+      modal: !modal
+    }));
+  };
 
-  togglePersistent = (): void =>
-    this.setState(
-      (prevState: PortalPageStateType): PortalPageStateType => ({
-        persistent: !prevState.persistent
-      })
-    );
+  togglePersistent = (): void => {
+    this.setState(({ persistent }: PortalPageStateType) => ({
+      persistent: !persistent
+    }));
+  };
 
   render() {
     return (
--- a/src/components/Checkbox.js
+++ b/src/components/Checkbox.js
@@ -116,7 +116,7 @@ class CheckboxComponent extends PureComponent<
       indeterminateProp !== undefined
         ? indeterminateProp
         : this.state.indeterminate;
-    // check for CheckMark or IndeterminateMark icons passed as props
+    // swap default checkmark for custom checkmarks if they exist
     const CheckMarkComponent = CheckMarkProp || CheckMark;
     const IndeterminateMarkComponent =
       indeterminateMarkProp || IndeterminateMark;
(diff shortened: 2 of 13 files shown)


ceciliamvrie:
NOTE: Since this pr was originally a flow typing pr, I don't really want to get rid of that work so I included some of it in this pr as well. I tried to keep the amount of lines changed to a minimum!

pages/bottomsheets.js
```diff
@@ -24,19 +24,19 @@ export default class PortalPage extends Component<
...
 
-  toggleModal = (): void =>
-    this.setState(
-      (prevState: PortalPageStateType): PortalPageStateType => ({
-        modal: !prevState.modal
-      })
-    );
+  toggleModal = (): void => {
+    this.setState((prevState: PortalPageStateType) => ({
+      ...prevState,
```
martahj: [Q] What was the reason for this change?

src/components/Checkbox.js
```diff
@@ -3,7 +3,7 @@ import React, { PureComponent } from "react";
 import styled from "styled-components";
 
 export type CheckboxPropsType = {|
-  CheckMark?: ?Object,
+  checkMark?: ?Object,
```
martahj: [Q] Since this will ultimately be aliased to an upper camel case component, why pass it as a lower case prop?

ceciliamvrie: Ah, but do we want to pas an upper camel case prop into the Checkbox component?

ceciliamvrie: It also might result to a breaking change potentially, if anyone happens to be using the `checkMark` prop right now

pages/bottomsheets.js
martahj: That seems.. surprising? I don't recall that being the case. Is there a setting we can toggle?

src/components/Checkbox.js
martahj: It would be a breaking change. We don't really have a policy around them, although we've made them in the past (and have a way of documenting them in PRs). It's probably worth opening an issue around how to handle breaking changes, such as when we would want to support both the new and old apis for a few releases (with deprecation warnings) vs when we would want to remove old apis.

ceciliamvrie:
@martahj Addressed comments!

martahj:
@gregorelee I think the dialog issue has been fixed in the latest push, could you check it out when you get a chance?

The two rows issue (might be the desired behavior, not sure) seems to exist on the [SMC website](https://styled-material-components.com/text-fields/) as well, so was not introduced in this PR. We also validated locally that the issue with the options was introduced in a previous PR.

PR #325, opened 2018-09-08
upgrade to sc4, fix bundle, more

brad-decker:
1. Upgrade styled-components to 4 @ beta
2. Refactor extend out
3. Refactor injectGlobal in favor of exporting OPTIONAL global style components
4. Add Rollup that creates two single file bundles
5. Make /pages use our built files instead of src to identify bundle issues sooner.

To do:
- [x] refactor tooltip to use refs correctly
- [x] update docs
- [x] update scripts
- [x] determine best development workflow using rollup + next
- [x] Test in production.


brad-decker:
1.5K of the diff is in the yarn.lock. most of the files come from changing three lines in all of the icons :(

brad-decker:
@gregorelee and @danieltisdale done. I made significant changes to the behind-the-scenes stuff with the menu. @gregorelee tag me when you get a minute today and i'll walk you through the expected behavior. @AriLFrankel @ceciliamvrie @martahj might want to tag you on this too make sure i didn't do anything stupid ;)

ceciliamvrie:
@brad-decker I'm encountering a bug with toggling back the persistent bottom sheet. It works fine when I open it, but it covers 2/3rds of the browser when I try to close it.

brad-decker:
@ceciliamvrie thanks! Resolved this one too

src/enhancers/withRipple.js
```diff
@@ -9,7 +9,7 @@ const rippleAnimation = keyframes`
 `;
 
 const withRipple = (ComposedComponent) => {
-  const RippledComponent = ComposedComponent.extend`
+  const RippledComponent = styled(ComposedComponent)`
```
petegivens: Were these changes performed by the SC automatic codemod? Looks like we're missing about 900 icons =x

pages/_app.js
```diff
@@ -0,0 +1,26 @@
...
+    }
+
+    return { pageProps };
+  }
+
+  render() {
+    const { Component, pageProps } = this.props;
+    return (
+      <Container>
+        <SMCGlobalStyles />
```
petegivens: can these be edited anywhere in the tree by importing `SMCGlobalStyles` and overwriting values in the css block?

src/enhancers/withRipple.js
brad-decker: what do you mean by missing? I did this using regex.

pages/_app.js
brad-decker: these are full on styled components, can take props, i'm not sure about using styled() or createGlobalStyles to extend them

src/enhancers/withRipple.js
petegivens: diff is showing that only 43 icon files were changed. shouldn't it be all of them? 
image

brad-decker: if you're using that tree addon it can only load 100 files ;)

petegivens: nothing to see here, folks.

PR #332, opened 2018-11-14
feat(Menu): add `fullWidth` prop

petegivens:
### Summary of changes:
 - add `fullWidth` prop to Menu component
  - if `true`, the width of the menu will be locked to the width of its anchor element.

### DIFF
--- a/src/components/Menu/Menu.js
+++ b/src/components/Menu/Menu.js
@@ -25,13 +25,15 @@ class MenuComponent extends Component {
       this.menu.current.contains(event.target) ||
       this.props.anchorEl === event.target ||
       isDescendant(this.props.anchorEl, event.target)
-    ) return;
+    ) {
+      return;
+    }
     this.props.onClose && this.props.onClose(event);
   };
 
   recalculatePosition = () => {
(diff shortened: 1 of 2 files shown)


src/components/Menu/Menu.js
```diff
@@ -49,27 +51,30 @@ class MenuComponent extends Component {
     if (openLeft) {
       position.left -= menuWidth + 5;
     }
-    if (attachBottom) position.top += anchorEl.offsetHeight + 5;
+    if (attachBottom) position.top += anchorEl.offsetHeight;
```
AriLFrankel: [q] acn you explain this and [this](https://github.com/MerlinLabs/styled-material-components/pull/332/files#diff-c46d8ddb9464108a35b17099c084ea07R69)?

petegivens: @AriLFrankel not sure where the + 5 came from, i just noticed that there was a 5px gap between the anchor element and the top of the menu so I removed it. I think menus should butt up immediately against the bottom of their anchor elements based on this image 
image

PR #339, opened 2019-01-23
Unfocus first MenuList item on page load

### LINKED ISSUE #284: TabIndex in Menu
The way that tabIndices are implemented in the Menu component leads to awkward focus behavior:
Open the menu and the first item is in hover state. Mousing through other options the first option remains in hover state

Removing the tabIndex would resolve this issue but be counterproductive to our efforts to increase accessibility in SMC

Joshawesome12:
### Summary of changes:
 - This will unfocus the first MenuList item that had been thought to be stuck in hover state. It was actually automatically focused by the line of code I removed. 

### Behaviors that should be QA'd:
 - Tab/Arrow Key functionality when switching focused items to make sure it works as desired/expected. 

### Github Issues closed by this PR:
 - Resolves #284

### DIFF
--- a/src/components/Menu/MenuList.js
+++ b/src/components/Menu/MenuList.js
@@ -2,7 +2,6 @@ import React from 'react';
 import styled from 'styled-components';
 
 class MenuListComponent extends React.Component {
-  
   menuList = React.createRef();
 
   handleKeyDown = (event) => {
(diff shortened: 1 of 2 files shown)


src/components/Menu/MenuList.js
```diff
@@ -1,10 +1,9 @@
 import React from 'react';
-import { findDOMNode } from 'react-dom';
 import styled from 'styled-components';
 
 class MenuListComponent extends React.Component {
   componentDidMount() {
-    findDOMNode(this.menuList.current).firstChild.focus();
+
```
AriLFrankel: please remove componentDidMount entirely instead of leaving it in as an empty function

